PR #43, opened 2025-07-30
feat: wipe all local data

lourou:
On top of account deletion, wipe all the databases and keys from the app group


Convos/Conversations List/ConversationsView.swift
```diff
@@ -53,6 +53,10 @@ struct ConversationsView: View {
                             }
                         }
 
+                        Button("Wipe all local data", role: .destructive) {
```
jarodl, line 56: Great! I would just make this the "Explode" button for now

PR #111, opened 2025-09-08
Trim whitespace from display/conversation names, skip empty updates

lourou:


Convos/Conversation Detail/ConversationViewModel.swift
```diff
@@ -33,15 +33,22 @@ class ConversationViewModel {
...
+            }
         }
     }
     var messages: [AnyMessage]
     var invite: Invite = .empty
     private(set) var profile: Profile = .empty(inboxId: "") {
         didSet {
-            displayName = profile.name ?? ""
+            // Don't update displayName here - it should only change when user edits it
+            // The profile update from the repository shouldn't overwrite user's input
```
jarodl, line 51: can probably just remove `didSet` entirely then

Convos/Shared Views/QuickEditView.swift
```diff
@@ -34,16 +35,16 @@ struct QuickEditView: View {
...
                 textFieldDelegate.action = onSubmit
                 textField.delegate = textFieldDelegate
+                textField.convos_setMaxLength(NameLimits.maxDisplayNameLength)
             }
             .padding(.horizontal, 16.0)
             .font(.system(size: 17.0))
             .tint(.colorTextPrimary)
             .foregroundStyle(.colorTextPrimary)
             .multilineTextAlignment(.center)
-            .textInputAutocapitalization(.words)
```
jarodl, line 43: ?

Convos/Conversation Detail/ConversationInfoEditView.swift
```diff
@@ -35,8 +37,12 @@ struct ConversationInfoEditView: View {
 
                 Section {
                     TextField(viewModel.conversationNamePlaceholder, text: $viewModel.conversationName)
-                        .textInputAutocapitalization(.words)
                         .lineLimit(1)
+                        .truncationMode(.tail)
+                        .frame(maxWidth: 166.0)
+                        .introspect(.textField, on: .iOS(.v26)) { textField in
+                            textField.convos_setMaxLength(NameLimits.maxConversationNameLength)
```
jarodl, line 44: can we do this validation in the view model?

jarodl (approved):

Convos/Conversation Detail/ConversationInfoEditView.swift
jarodl: using `onChange` is a better approach than `introspect`, I think we only use the introspect library in one place and it'd be good to keep it that way, we don't want things to break in future iOS versions

jarodl (changes requested):

Convos/Shared Views/QuickEditView.swift
lourou: Capitalizing on every word seems odd to me, I end up removing the cap tab for every new word, especially on smaller words like 2-3 letters, they shouldn't be capitalized I think

PR #152, opened 2025-09-23
Integrate backend invite acceptance flow with invite code use tracking

lourou:
Replaces DM-based invite flow with API-driven acceptance that tracks invite code usage.

Tracking invite code usage provides a temporary audit trail (until the invite code is deleted) of who invited each member and enables permission checks for group metadata updates based on the invitation hierarchy.

Changes:
  - Added `acceptRequestToJoin()` which implements a new backend endpoint to accept join requests and create `InviteCodeUse` records
  - NSE now calls backend first to accept requests, ensuring code use is tracked before adding to XMTP
  - Removed `InviteJoinRequestsManager` that monitored DMs for invite codes
  - Backend automatically cleans up accepted requests
  - Added error handling and idempotency checks
  - Attempting to fix the double join issue that surfaced


Convos/Conversation Creation/NewConversationViewModel.swift
```diff
@@ -124,6 +125,9 @@ class NewConversationViewModel: Identifiable {
     }
 
     func join(inviteUrlString: String) -> Bool {
+        Logger.info("🔍 [JOIN] Called with: \(inviteUrlString)")
```
jarodl, line 128: I know this is nit-picky, but can we remove emojis and things like [JOIN] from the logs? `Logger.info` already adds the method signature and I think having the emojis make some logs stand out more than others, which is useful if you're currently debugging that flow, but if not makes them all a bit harder to parse

jarodl (changes requested):

PR #168, opened 2025-10-15
v2 backend integration

lourou:
### Integrate v2 backend by replacing push registration, authentication, and invite flows across `ConvosCore` and app surfaces, including `ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift` and `ConvosCore/Sources/ConvosCore/Inboxes/InboxStateMachine.swift`
This pull request adopts the v2 backend across notification, authentication, invite, and storage paths, refactors state managers and protocols to use `clientId` and protocol-oriented dependencies, and introduces Protobuf-based invite and custom metadata handling. It centralizes device registration with AppCheck, updates push processing to route protocol messages uniformly, and reworks conversation creation/join flows to use signed invite slugs and client-bound topic subscriptions.

- Replace v1 APIs with v2 endpoints for device registration, auth, and topic subscriptions in `ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift` and add `ConvosCore/Sources/ConvosCore/API/ConvosAPIClient+Models.swift` models; introduce `EmptyResponse` and AppCheck header usage.

- Refactor `InboxStateMachine` to carry `inboxId` through states, persist inboxes with `clientId`, register devices via `DeviceRegistrationManager`, and remove backend registration flows; see `ConvosCore/Sources/ConvosCore/Inboxes/InboxStateMachine.swift`.

- Convert push notification payloads and handlers to `clientId`-based matching, drop `notificationType`, and route handling through protocol message processing; see `ConvosCore/Sources/ConvosCore/Notifications/PushNotificationPayload.swift` and `ConvosCore/Sources/ConvosCore/Inboxes/CachedPushNotificationHandler.swift`.

- Implement signed invite generation/validation using SwiftProtobuf and secp256k1, add URL-safe slug helpers, and cryptographic conversation tokens; see `ConvosCore/Sources/ConvosCore/Protobuf Models/InviteProtobufExtensions.swift` and `ConvosCore/Sources/ConvosCore/Protobuf Models/InviteCode.swift`.

- Rework `ConversationStateMachine` to validate invite codes before joining, subscribe using `clientId`, and emit origins; add identity-aware writers and cleanup; see `ConvosCore/Sources/ConvosCore/Inboxes/ConversationStateMachine.swift`.

- Introduce `DeviceRegistrationManager` and static `PushNotificationRegistrar` utility; remove keychain/APNS backend registration from previous paths; see `ConvosCore/Sources/ConvosCore/Device/DeviceRegistrationManager.swift` and `ConvosCore/Sources/ConvosCore/Inboxes/PushNotificationRegistrar.swift`.

- Update storage schema and writers to include `clientId`, `inviteTag`, composite keys, and draft placeholders; change database file to `convos_v2.sqlite`; see `ConvosCore/Sources/ConvosCore/Storage/SharedDatabaseMigrator.swift` and `ConvosCore/Sources/ConvosCore/Storage/DatabaseManager.swift`.

- Replace group-prefixed metadata/permissions with conversation-prefixed APIs and inject `InviteWriter`; see `ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationMetadataWriter.swift` and `ConvosCore/Sources/ConvosCore/Storage/Repositories/ConversationPermissionsRepository.swift`.

- Revise `SessionManager` to be a class using a `DispatchQueue`, make service access synchronous, and add `inboxId(for:)`; see `ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift`.

- Update UI flows to new join semantics, invite URL, and bottom bar injection; see `Convos/Conversation Creation/NewConversationView.swift` and `Convos/Conversation Detail/ConversationView.swift`.

#### 📍Where to Start
Start with the v2 API integration and device registration in `ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift` to understand endpoint changes, then follow the notification and state flow through `ConvosCore/Sources/ConvosCore/Inboxes/InboxStateMachine.swift`.

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 0942273. 22 files reviewed, 41 issues evaluated, 32 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>Convos/Conversation Detail/ConversationShareView.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 25](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/Convos/Conversation Detail/ConversationShareView.swift#L25): Unvalidated share item may be empty or invalid. `ConversationShareView` passes `items: [invite.inviteURLString]` to `AutoShareSheetView` (line 25) without any guard for validity or non-emptiness, while the QR code content is conditionally shown only if `invite.inviteURL` exists (lines 61–67). If `invite.inviteURL` is `nil`, it's plausible that `invite.inviteURLString` is empty or otherwise invalid, resulting in a share sheet with a malformed or empty item. This produces an externally visible artifact that may be rejected or behave unpredictably across share targets. To preserve contract parity, the share sheet items should be derived from the same validated source as the QR code (e.g., pass `invite.inviteURL` when present, or provide a clear fallback/error path when it's absent). <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/ConvosAppDelegate.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 24](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/Convos/ConvosAppDelegate.swift#L24): Push token change notification is no longer posted unconditionally from `ConvosAppDelegate.application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`. The delegate previously posted `NotificationCenter` event `.convosPushTokenDidChange` on every registration attempt; now it delegates to `PushNotificationRegistrar.save(token:)`, which posts only when the token differs from the last saved value. If any components rely on receiving the event even when the token is unchanged (e.g., to re-sync on app launch), they will miss it, changing externally visible behavior and potentially causing the backend token sync to be skipped. <b>[ Low confidence ]</b>
- [line 62](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/Convos/ConvosAppDelegate.swift#L62): Explosion notification path posts `NotificationCenter` event with potentially empty `inboxId` and/or `conversationId`. In the explosion branch, the code only checks for the existence of the keys and type-casts to `String`, but does not validate non-emptiness before posting. This is inconsistent with the regular notification path, which guards `conversationId` non-emptiness. Emitting empty IDs can cause downstream consumers to fail or mis-route. Add explicit `!isEmpty` checks for `inboxId` and `conversationId` before posting, mirroring the regular path. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Debug View/DebugView.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 158](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/Convos/Debug View/DebugView.swift#L158): The `ShareLink` is shown whenever `debugFileURL` is non-nil, but `prepareDebugInfoFile()` sets `debugFileURL` unconditionally even if writing the file fails due to the `try?` at `debugInfo.write(...)`. This can result in `ShareLink(item: debugFileURL)` attempting to share a non-existent or empty file, causing user-visible failure and violating the external artifact validity contract. A safer approach is to only set `debugFileURL` after a successful write, or additionally check that the file exists before rendering the `ShareLink` (e.g., `FileManager.default.fileExists(atPath: debugFileURL.path)`). <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/DeepLinking/DeepLinkHandler.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 20](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/Convos/DeepLinking/DeepLinkHandler.swift#L20): Deep link handling now rejects previously valid universal link formats that used a path component (e.g., `https://domain.com/invite-code`). `DeepLinkHandler.destination(for:)` depends on `url.convosInviteCode`, which was changed to only accept `https://.../v2?i=invite-code` or `convos://join/invite-code`. If upstream still emits the old path-based format, `destination(for:)` will return `nil` and dismiss the deep link as missing an invite code, changing externally visible behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Device/DeviceRegistrationManager.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 78](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Device/DeviceRegistrationManager.swift#L78): `registerDeviceIfNeeded()` treats any non-`nil` `pushToken` as valid, including an empty string. If `PushNotificationRegistrar.token` can ever be `""` (empty), this path will: (1) attempt registration with an invalid token value; (2) persist `hasRegisteredDevice` and `lastRegisteredDevicePushToken` using the empty string; and (3) subsequently skip re-registration until the token changes from `""`, potentially leaving the device considered "registered with push token" when the server rejects or ignores the empty token. The code should validate that `pushToken` is non-empty (e.g., trim and check for `isEmpty`) and treat empty tokens as `nil` to avoid sending/recording invalid tokens and to continue retrying until a valid token is present. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift — 0 comments posted, 3 evaluated, 2 filtered</summary>

- [line 50](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L50): `XMTPiOS.Group.updateInviteTag()` performs a non-atomic sequence with external effects: it retrieves current metadata, generates a new tag, and calls `updateDescription(...)` to persist it. If the secure random generation (`SecRandomCopyBytes`) or `updateDescription` fails, the upstream conversation creation flow (as shown in example usage) will abort after publishing the conversation, leaving a partially created remote conversation without updated metadata and without any compensating cleanup. This risks inconsistent system state and violates the single paired cleanup and atomicity guidance for sequences with external effects. <b>[ Invalidated by documentation search ]</b>
- [line 111](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L111): `XMTPiOS.Group.updateProfile(_:)` does not validate that the provided `MemberProfile` belongs to the same conversation (`profile.conversationId` vs `self.id`). The upsert logic uses only `inboxID` for matching, enabling a caller to mistakenly apply a profile from another conversation, corrupting the group’s custom metadata. A guard should enforce that `profile.conversationId == id` before upserting. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Protobuf Models/InviteCode.swift — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 163](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/InviteCode.swift#L163): `deriveKey(privateKey:inboxId:)` accepts any non-empty `privateKey` length and derives a symmetric key via HKDF. Elsewhere, the signing function (`sign(with:)`) requires a strict 32-byte secp256k1 private key. This inconsistent key-length validation across the module can lead to runtime surprises and decryption failures or weak/incorrect key derivation if a caller passes a non-32-byte key to `makeConversationToken`/`decodeConversationToken`. At minimum, this is a contract asymmetry within the same feature area that should be aligned to avoid misuse. <b>[ Low confidence ]</b>
- [line 227](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/InviteCode.swift#L227): `packConversationId(_:)` forbids empty conversation IDs (`guard !id.isEmpty else { throw Error.emptyConversationId }`), but the implementation and comments in the UTF-8 branch explicitly describe supporting a zero-length string using a 0 sentinel followed by a 2-byte big-endian length. This contradiction between implementation and the documented format creates uncertainty about whether empty IDs are allowed, and can cause decode/encode asymmetry if other components produce empty-string tokens (decode path would accept and return empty strings, while encode path cannot produce them). <b>[ Low confidence ]</b>
- [line 260](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/InviteCode.swift#L260): `unpackConversationId(_:)` returns a lowercased UUID string via `uuid.uuidString.lowercased()`. This changes the original UUID’s case and can break equality checks or downstream lookups that expect the original representation. It also contradicts the decode function’s contract to "Recover the original conversationId". If the original `conversationId` was stored/persisted in uppercase (Swift's `UUID.uuidString` is typically uppercase), decoding will yield a different string, potentially causing mismatches (e.g., not finding a conversation by ID). <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Protobuf Models/InviteProtobufExtensions.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 149](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/InviteProtobufExtensions.swift#L149): `SignedInvite.toURLSafeSlug()` can theoretically return an empty string if `serializedData()` is empty, while `SignedInvite.fromURLSafeSlug(_:)` uses `base64URLDecoded()` which throws on empty input. This mirrors the metadata parity issue and can cause a failure to decode an empty slug. Even if empty invites are unlikely, the encoder/decoder should have consistent handling for empty cases. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Protobuf Models/proto/invite.proto — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 6](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/proto/invite.proto#L6): In `proto3`, scalar fields without `optional` (e.g., `string conversationToken`, `string creator_inbox_id`, `string tag`) do not have presence and default to empty strings when omitted. This makes it impossible at runtime to distinguish between "not provided" and "explicitly provided as empty", which can silently accept malformed or incomplete `InvitePayload` instances if downstream logic expects these to be present (i.e., non-empty). To preserve presence and allow validation for required-like fields, consider using `optional string` (with presence) or moving presence enforcement to a message-level validation layer and documenting that empty values are rejected. As-is, serialized messages can omit these fields and still parse to empty strings, producing ambiguous and potentially invalid artifacts. <b>[ Low confidence ]</b>
- [line 26](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Protobuf Models/proto/invite.proto#L26): `bytes signature` in `SignedInvite` is unconstrained and has no presence requirement or validation in the schema. In proto3, if the field is omitted, it defaults to empty bytes at runtime, which is indistinguishable from an explicitly provided empty signature. This permits emitting or accepting a `SignedInvite` that is effectively unsigned (empty `signature`), potentially violating cryptographic invariants. If a non-empty signature is required, consider using `optional bytes` (to detect omission) and enforce non-empty length at the application level, or document and validate minimum length (e.g., 65 bytes) downstream. As-is, malformed invites can parse without error and propagate. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationStateManager.swift — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 47](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationStateManager.swift#L47): `waitForConversationReadyResult(timeout:)` ignores the provided `timeout` and returns immediately with `.created` regardless of the actual state. This violates contract parity with a realistic implementation that would wait for readiness, respect timeout, and return the actual origin. Callers relying on waiting semantics may proceed prematurely, causing operations to run before the conversation is ready. <b>[ Low confidence ]</b>
- [line 54](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationStateManager.swift#L54): `addObserver(_:)` allows the same `ConversationStateObserver` instance to be added multiple times without any deduplication or guard, which can result in duplicated notifications for the same state change. This violates uniqueness constraints for observer registration and can cause handlers to run more than once per event. <b>[ Low confidence ]</b>
- [line 113](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationStateManager.swift#L113): The `MockConversationStateManager`'s observer management (`addObserver`, `removeObserver`, and `notifyObservers`) is not thread-safe. The `observers` array is mutated from multiple async methods (`addObserver`, `removeObserver`, `createConversation`, `joinConversation`, `delete`) without any synchronization or actor isolation, which can lead to data races, missed notifications, or crashes when concurrently accessed. Additionally, `notifyObservers` rebuilds the `observers` array while invoking callbacks, and those callbacks can themselves call `removeObserver`, causing re-entrancy and concurrent mutation of the same collection. This violates the at-most-once and atomic consumption guarantees required for callbacks and state transitions. <b>[ Invalidated by documentation search ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationStateManager.swift — 0 comments posted, 4 evaluated, 4 filtered</summary>

- [line 168](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationStateManager.swift#L168): `ConversationStateManager.handleStateChange(_:)` sets `isReady = true` for both `.ready` and `.joining` states via the combined pattern `case .ready(let result), .joining(invite: _, placeholder: let result):`. The `.joining` state (per upstream UI logic in `NewConversationViewModel.handleStateChange`) represents a waiting state where the inviter must accept; it should not be treated as fully ready. Marking `.joining` as ready prematurely enables ready-dependent behaviors (e.g., `ConversationStateAwareComponent.isConversationReady` will return `true`, potentially enabling UI actions or flows that require a fully ready conversation). This is a contract mismatch and can lead to incorrect runtime behavior and inconsistent UI state. <b>[ Low confidence ]</b>
- [line 192](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationStateManager.swift#L192): `ConversationStateManager.addObserver(_:)` does not enforce uniqueness and can add the same `observer` instance multiple times. This causes multiple notifications per state change to the same object, violating at-most-once semantics for callbacks and potentially leading to duplicated side effects. If duplicates are not desired, this is a runtime bug. A guard to prevent duplicates or a set-like structure keyed by object identity should be used. <b>[ Low confidence ]</b>
- [line 251](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationStateManager.swift#L251): `ConversationStateManager.joinConversation(inviteCode:)` returns immediately after enqueuing a validation/join via `stateMachine.join(inviteCode:)` without awaiting any validation completion. The comment `// @jarodl This should wait for validation, but not readiness` indicates the intended behavior. As written, callers may treat successful return as a validated join, leading to premature success handling (see `NewConversationViewModel.joinConversation` calling `handleJoinSuccess()` right after). This creates a runtime contract mismatch and can cause incorrect UI flow or side effects before validation completes. <b>[ Low confidence ]</b>
- [line 313](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationStateManager.swift#L313): `ConversationStateAwareComponent`’s observer closure spawns a new `Task { @MainActor in ... }` for every state change, even though `notifyObservers` calls observers on the main actor. This extra Task hop can introduce reordering of state updates under load (tasks scheduled may not execute strictly FIFO with respect to synchronous main-actor calls), causing out-of-order or stale `state` assignments and `conversationStateDidChange(_:)` notifications. The component should update directly on the main actor without creating a new Task, or otherwise ensure ordered delivery (e.g., using `MainActor.run` inline). <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosTests/ConversationStateManagerTests.swift — 0 comments posted, 7 evaluated, 7 filtered</summary>

- [line 52](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L52): When inbox registration times out, `registerTemporaryInboxId()` returns an empty `String` (""), logs an issue, and continues. Callers (e.g., `testSendingMessageCreatesConversation`) then use this empty `inboxId` to construct a `MemberProfile` without validation. An empty inbox ID is an invalid identifier and can lead to inconsistent conversation state, nondeterministic behavior, or silent failures downstream. The function should fail fast (e.g., by throwing) rather than returning an empty string, and callers should validate that a non-empty inbox ID was obtained before proceeding. <b>[ Test / Mock code ]</b>
- [line 147](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L147): The test uses `guard` to handle preconditions (inboxes available and messaging service readiness) and then `return`s without throwing or asserting, which can allow the test to exit early and pass despite the failure. Specifically, if no inboxes are found or the messaging service doesn't become ready, the test only logs and `Issue.record(...)` but does not fail the test. Unless `Issue.record` integrates with the test runner to fail the test, this creates false positives where the test appears successful. This occurs at the inbox check (`guard let firstInboxes = firstInboxes, let inbox = firstInboxes.first else { ... return }`) and at the readiness check (`guard let _ = readyResult else { ... return }`). The test should explicitly fail (e.g., by throwing or using `#expect(false)`), not silently return. <b>[ Test / Mock code ]</b>
- [line 183](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L183): The test relies on `waitForFirstMatch(where:)` with its default timeout (2 seconds) for multiple asynchronous state transitions, which can cause flaky failures in slower environments. The calls at waiting for DM state (`conversationPublisher.waitForFirstMatch(where: { $0?.kind == .dm })`), for group state (`... { $0?.kind == .group }`), and again for DM state after removal all use the default timeout. Given that upstream publishers and state transitions can take longer (e.g., messaging service setup just waited with a 30s timeout), using a much shorter default can result in intermittent timeouts and false test failures. Each of these waits should specify a sufficiently large timeout consistent with the environment or leverage readiness signals. <b>[ Test / Mock code ]</b>
- [line 256](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L256): `testSendingMessageCreatesConversation()` does not validate that `otherInboxId` is non-empty after calling `registerTemporaryInboxId()`. If registration times out, `registerTemporaryInboxId()` returns "", and the test proceeds to construct `MemberProfile(inboxId: otherInboxId, ...)` with an empty ID, which is an invalid identifier and can cause downstream inconsistencies or false positives/negatives. Add a guard to ensure `otherInboxId` is non-empty and fail the test early if not. <b>[ Test / Mock code ]</b>
- [line 261](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L261): The test launches an unstructured `Task` to call `writer.add(profile:)` and does not `await` its completion or handle errors. If `add` throws or is slow, the test concurrently races to wait for the conversation state, which can cause nondeterministic behavior, timeouts, or swallowed errors. Use `try await writer.add(profile:)` directly (or `await` the `Task` result) to ensure ordering and reliable failure propagation. <b>[ Test / Mock code ]</b>
- [line 271](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L271): The test launches an unstructured `Task` to call `writer.send(text:)` and does not `await` its completion or handle errors. If `send` throws or is delayed, the test may proceed to wait for messages before the send completes, causing nondeterministic behavior or timeouts, and errors are silently ignored. Use `try await writer.send(text:)` directly (or capture and await the `Task` result) to ensure reliable sequencing and error propagation. <b>[ Test / Mock code ]</b>
- [line 342](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/ConvosTests/ConversationStateManagerTests.swift#L342): The test creates a `MemberProfile` using `otherInboxId` returned from `registerTemporaryInboxId()` without validating that a non-empty inbox ID was obtained. The helper `registerTemporaryInboxId()` returns an empty string (`""`) on failure rather than throwing, which is reachable under timeout or registration failure. Proceeding with `MemberProfile(inboxId: "")` can yield invalid state, cause matching to fail, or make the subsequent waits (`waitForFirstMatch`) hang or flake. The test should guard that `otherInboxId` is non-empty (and fail early) or change `registerTemporaryInboxId()` to throw on failure so callers must handle it. <b>[ Test / Mock code ]</b>
</details>

<details>
<summary>NotificationService/NotificationService.swift — 0 comments posted, 4 evaluated, 3 filtered</summary>

- [line 67](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/NotificationService/NotificationService.swift#L67): Using `[weak self]` and `guard let self = self else { return }` inside the `Task` can result in silently abandoning processing without invoking `contentHandler` if `self` is deallocated during processing. This leaves the request without a defined terminal state. While avoiding retain cycles is good, for a notification service extension you typically must guarantee delivery exactly once. If `self` becomes nil, neither the processed content nor a fallback empty content is delivered from this path. <b>[ Low confidence ]</b>
- [line 88](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/NotificationService/NotificationService.swift#L88): Possible double invocation of `contentHandler` due to unsynchronized access from concurrent paths. Both the processing `Task` and `serviceExtensionTimeWillExpire()` read and invoke `self.contentHandler` without any synchronization or atomic swap-and-clear. If the task completes around the same time the system calls `serviceExtensionTimeWillExpire`, both paths can observe a non-nil `contentHandler` and invoke it, violating the at-most-once callback guarantee. This race exists at both delivery sites in `didReceive` (lines where `handler(self.contentHandler)` is invoked) and the timeout handler. <b>[ Invalidated by documentation search ]</b>
- [line 101](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/NotificationService/NotificationService.swift#L101): Cancellation path may fail to deliver any content, leaving the request without a defined terminal outcome. In the `catch is CancellationError` block, the code logs and deliberately does not invoke `contentHandler`, relying on `serviceExtensionTimeWillExpire` to deliver content. If cancellation is triggered by `currentProcessingTask?.cancel()` when a new request arrives (not due to timeout), `serviceExtensionTimeWillExpire` may not be called, and the original request's content handler will never be invoked. This can result in the system delivering the original notification unintentionally or cause inconsistent suppression behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Scripts/setup.sh — 0 comments posted, 4 evaluated, 1 filtered</summary>

- [line 33](https://github.com/ephemeraHQ/convos-ios/blob/094227397c8d85deaaafacc32296bdbcc90e1ecf/Scripts/setup.sh#L33): Hook installation only runs when `hooks/*` are not symlinks (`[ ! -L hooks/... ]`). If a hook already exists as a symlink but points to the wrong or stale target, it will not be updated, leaving a broken setup. Given the intent is to ensure hooks are installed/updated, this should force-update symlinks unconditionally or verify the symlink target and relink if mismatched. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/Inboxes/ConversationStateMachine.swift
+++ b/ConvosCore/Sources/ConvosCore/Inboxes/ConversationStateMachine.swift
@@ -65,6 +65,7 @@ public actor ConversationStateMachine {
     private let inboxStateManager: any InboxStateManagerProtocol
     private let databaseReader: any DatabaseReader
     private let databaseWriter: any DatabaseWriter
+    private let environment: AppEnvironment
 
     private var currentTask: Task<Void, Never>?
     private var streamConversationsTask: Task<Void, Never>?
@@ -126,11 +127,13 @@ public actor ConversationStateMachine {
         identityStore: any KeychainIdentityStoreProtocol,
         databaseReader: any DatabaseReader,
         databaseWriter: any DatabaseWriter,
+        environment: AppEnvironment
     ) {
         self.inboxStateManager = inboxStateManager
         self.identityStore = identityStore
         self.databaseReader = databaseReader
         self.databaseWriter = databaseWriter
+        self.environment = environment
     }
 
     deinit {
@@ -350,6 +353,15 @@ public actor ConversationStateMachine {
             conversationId: externalConversationId,
             origin: .created
         )))
+
+        // Clear unused inbox from keychain now that conversation is successfully created
+        await UnusedInboxCache.shared
+            .clearUnusedInbox(
+                with: client.inboxId,
+                databaseWriter: databaseWriter,
+                databaseReader: databaseReader,
+                environment: environment
+            )
     }
 
     private func handleValidate(inviteCode: String, previousResult: ConversationReadyResult?) async throws {
@@ -508,6 +520,15 @@ public actor ConversationStateMachine {
                         conversationId: conversation.id,
                         origin: .joined
                     )))
+
+                    // Clear unused inbox from keychain now that conversation is successfully joined
+                    await UnusedInboxCache.shared
+                        .clearUnusedInbox(
+                            with: client.inboxId,
+                            databaseWriter: databaseWriter,
+                            databaseReader: databaseReader,
+                            environment: environment
+                        )
 
                     // Clean up previous conversation if it exists and is different
                     await self.cleanUpPreviousConversationIfNeeded(
--- a/Convos/Conversation Creation/NewConversationViewModel.swift
+++ b/Convos/Conversation Creation/NewConversationViewModel.swift
@@ -42,31 +42,6 @@ class NewConversationViewModel: Identifiable {
     private(set) var currentError: Error?
     private(set) var conversationState: ConversationStateMachine.State = .uninitialized
 
-    // MARK: - Computed Properties
-
-    /// Whether the conversation is in a loading/processing state
-    var isProcessing: Bool {
-        isCreatingConversation || isValidatingInvite || isWaitingForInviteAcceptance
-    }
-
-    /// Whether there is an active error
-    var hasError: Bool {
-        currentError != nil
-    }
-
-    /// Localized error message for display
-    var errorMessage: String? {
-        currentError?.localizedDescription
-    }
-
-    /// Whether the conversation is ready for use
-    var isConversationReady: Bool {
-        if case .ready = conversationState {
-            return true
-        }
-        return false
-    }
-
     // MARK: - Private
 
     private let conversationStateManager: any ConversationStateManagerProtocol
@@ -189,19 +164,6 @@ class NewConversationViewModel: Identifiable {
                 Logger.error("Failed deleting conversation: \(error.localizedDescription)")
             }
         }
-    }
-
-    func clearError() {
-        currentError = nil
-    }
-
-    func retryAfterError() {
-        guard let error = currentError else { return }
-        Logger.info("Retrying after error: \(error.localizedDescription)")
-        currentError = nil
-
-        // If we were in the middle of joining, could potentially retry
-        // For now, just clear the error and let the user try again
     }
 
     // MARK: - Private
--- a/Convos/Conversations List/ConversationsViewModel.swift
+++ b/Convos/Conversations List/ConversationsViewModel.swift
@@ -56,7 +56,7 @@ final class ConversationsViewModel {
                 hasCreatedMoreThanOneConvo = true
             }
 
-            hasEarlyAccess = true// conversationsCount > 0
+            hasEarlyAccess = conversationsCount > 0
         }
     }
 
@@ -102,6 +102,7 @@ final class ConversationsViewModel {
     private var localStateWriters: [String: any ConversationLocalStateWriterProtocol] = [:]
     private var cancellables: Set<AnyCancellable> = .init()
     private var leftConversationObserver: Any?
+    private var newConversationViewModelTask: Task<Void, Never>?
 
     init(session: any SessionManagerProtocol) {
         self.session = session
@@ -123,7 +124,8 @@ final class ConversationsViewModel {
             self.hasEarlyAccess = false
         }
         if !hasEarlyAccess {
-            Task {
+            newConversationViewModelTask = Task { [weak self] in
+                guard let self else { return }
                 self.newConversationViewModel = await NewConversationViewModel.create(
                     session: session,
                     showingFullScreenScanner: true,
@@ -135,6 +137,7 @@ final class ConversationsViewModel {
     }
 
     deinit {
+        newConversationViewModelTask?.cancel()
         if let leftConversationObserver {
             NotificationCenter.default.removeObserver(leftConversationObserver)
         }
@@ -157,7 +160,9 @@ final class ConversationsViewModel {
             presentingMaxNumberOfConvosReachedInfo = true
             return
         }
-        Task {
+        newConversationViewModelTask?.cancel()
+        newConversationViewModelTask = Task { [weak self] in
+            guard let self else { return }
             newConversationViewModel = await NewConversationViewModel.create(
                 session: session,
                 autoCreateConversation: true,
@@ -171,7 +176,9 @@ final class ConversationsViewModel {
             presentingMaxNumberOfConvosReachedInfo = true
             return
         }
-        Task {
+        newConversationViewModelTask?.cancel()
+        newConversationViewModelTask = Task { [weak self] in
+            guard let self else { return }
             newConversationViewModel = await NewConversationViewModel.create(
                 session: session,
                 showingFullScreenScanner: true,
@@ -192,7 +199,9 @@ final class ConversationsViewModel {
             presentingMaxNumberOfConvosReachedInfo = true
             return
         }
-        Task {
+        newConversationViewModelTask?.cancel()
+        newConversationViewModelTask = Task { [weak self] in
+            guard let self else { return }
             newConversationViewModel = await NewConversationViewModel.create(
                 session: session,
                 delegate: self
@@ -209,7 +218,7 @@ final class ConversationsViewModel {
                 try await session.deleteAllInboxes()
 
                 // Clear all cached writers
-                self.localStateWriters.removeAll()
+                await MainActor.run { self.localStateWriters.removeAll() }
             } catch {
                 Logger.error("Error deleting all accounts: \(error)")
             }
@@ -227,7 +236,7 @@ final class ConversationsViewModel {
                 try await session.deleteInbox(inboxId: conversation.inboxId)
 
                 // Remove cached writer for deleted inbox
-                self.localStateWriters.removeValue(forKey: conversation.inboxId)
+                _ = await MainActor.run { self.localStateWriters.removeValue(forKey: conversation.inboxId) }
             } catch {
                 Logger.error("Error leaving convo: \(error.localizedDescription)")
             }
--- a/Convos/ConvosAppDelegate.swift
+++ b/Convos/ConvosAppDelegate.swift
@@ -20,7 +20,7 @@ class ConvosAppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCent
         let token = tokenParts.joined()
 
         Logger.info("Received device token from APNS: \(token)")
-        // Store token in shared storage (automatically notifies listeners)
+        // Store token in shared storage
         PushNotificationRegistrar.save(token: token)
         Logger.info("Stored device token in shared storage")
     }
--- a/ConvosCore/Sources/ConvosCore/Inboxes/PushNotificationRegistrar.swift
+++ b/ConvosCore/Sources/ConvosCore/Inboxes/PushNotificationRegistrar.swift
@@ -10,6 +10,9 @@ public final class PushNotificationRegistrar {
     /// Saves the push token to UserDefaults and notifies observers of the change.
     /// Called by AppDelegate when APNS token is received.
     public static func save(token: String) {
+        let existingToken = UserDefaults.standard.string(forKey: tokenKey)
+        guard token != existingToken else { return }
+
         UserDefaults.standard.set(token, forKey: tokenKey)
         NotificationCenter.default.post(name: .convosPushTokenDidChange, object: nil)
     }
--- a/ConvosCore/Sources/ConvosCore/Messaging/MessagingService.swift
+++ b/ConvosCore/Sources/ConvosCore/Messaging/MessagingService.swift
@@ -9,6 +9,7 @@ final class MessagingService: MessagingServiceProtocol {
     internal let identityStore: any KeychainIdentityStoreProtocol
     internal let databaseReader: any DatabaseReader
     internal let databaseWriter: any DatabaseWriter
+    private let environment: AppEnvironment
     private var cancellables: Set<AnyCancellable> = []
 
     static func authorizedMessagingService(
@@ -33,7 +34,8 @@ final class MessagingService: MessagingServiceProtocol {
             authorizationOperation: authorizationOperation,
             databaseWriter: databaseWriter,
             databaseReader: databaseReader,
-            identityStore: identityStore
+            identityStore: identityStore,
+            environment: environment
         )
     }
 
@@ -52,12 +54,14 @@ final class MessagingService: MessagingServiceProtocol {
     internal init(authorizationOperation: AuthorizeInboxOperation,
                   databaseWriter: any DatabaseWriter,
                   databaseReader: any DatabaseReader,
-                  identityStore: any KeychainIdentityStoreProtocol) {
+                  identityStore: any KeychainIdentityStoreProtocol,
+                  environment: AppEnvironment) {
         self.identityStore = identityStore
         self.authorizationOperation = authorizationOperation
         self.inboxStateManager = InboxStateManager(stateMachine: authorizationOperation.stateMachine)
         self.databaseReader = databaseReader
         self.databaseWriter = databaseWriter
+        self.environment = environment
     }
 
     deinit {
@@ -91,7 +95,8 @@ final class MessagingService: MessagingServiceProtocol {
             inboxStateManager: inboxStateManager,
             identityStore: identityStore,
             databaseReader: databaseReader,
-            databaseWriter: databaseWriter
+            databaseWriter: databaseWriter,
+            environment: environment
         )
     }
 
--- a/ConvosCore/Sources/ConvosCore/Messaging/UnusedInboxCache.swift
+++ b/ConvosCore/Sources/ConvosCore/Messaging/UnusedInboxCache.swift
@@ -26,7 +26,7 @@ actor UnusedInboxCache {
         environment: AppEnvironment
     ) async {
         // Check if we already have an unused messaging service ready
-        if unusedMessagingService != nil {
+        guard unusedMessagingService == nil else {
             Logger.debug("Unused messaging service already exists")
             return
         }
@@ -68,9 +68,6 @@ actor UnusedInboxCache {
             // Clear the reference
             unusedMessagingService = nil
 
-            // Clear from keychain
-            clearUnusedInboxFromKeychain()
-
             // Save to database now that it's being consumed by the user
             do {
                 let result = try await unusedService.inboxStateManager.waitForInboxReadyResult()
@@ -83,36 +80,14 @@ actor UnusedInboxCache {
                 Logger.error("Failed to save consumed inbox to database: \(error)")
             }
 
-            // Schedule creation of a new unused inbox for next time
-            Task(priority: .background) { [weak self] in
-                guard let self else { return }
-                await createNewUnusedInbox(
-                    databaseWriter: databaseWriter,
-                    databaseReader: databaseReader,
-                    environment: environment
-                )
-            }
+            await unusedService.registerForPushNotifications()
 
-            // Device registration happens automatically when inbox becomes ready
             return unusedService
         }
 
         // Check for an unused inbox ID in keychain (fallback)
         if let unusedInboxId = getUnusedInboxFromKeychain() {
             Logger.info("Using unused inbox ID from keychain: \(unusedInboxId)")
-
-            // Clear from keychain
-            clearUnusedInboxFromKeychain()
-
-            // Schedule creation of a new unused inbox for next time
-            Task(priority: .background) { [weak self] in
-                guard let self else { return }
-                await createNewUnusedInbox(
-                    databaseWriter: databaseWriter,
-                    databaseReader: databaseReader,
-                    environment: environment
-                )
-            }
 
             // Use the existing inbox with authorize
             // Note: The authorize flow in InboxStateMachine.handleAuthorize() will
@@ -130,7 +105,8 @@ actor UnusedInboxCache {
                 authorizationOperation: authorizationOperation,
                 databaseWriter: databaseWriter,
                 databaseReader: databaseReader,
-                identityStore: identityStore
+                identityStore: identityStore,
+                environment: environment
             )
         }
 
@@ -160,7 +136,8 @@ actor UnusedInboxCache {
             authorizationOperation: authorizationOperation,
             databaseWriter: databaseWriter,
             databaseReader: databaseReader,
-            identityStore: identityStore
+            identityStore: identityStore,
+            environment: environment
         )
     }
 
@@ -186,7 +163,8 @@ actor UnusedInboxCache {
             authorizationOperation: authorizationOperation,
             databaseWriter: databaseWriter,
             databaseReader: databaseReader,
-            identityStore: identityStore
+            identityStore: identityStore,
+            environment: environment
         )
 
         do {
@@ -249,7 +227,8 @@ actor UnusedInboxCache {
             authorizationOperation: authorizationOperation,
             databaseWriter: databaseWriter,
             databaseReader: databaseReader,
-            identityStore: identityStore
+            identityStore: identityStore,
+            environment: environment
         )
 
         do {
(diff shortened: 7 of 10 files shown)


ConvosCore/Sources/ConvosCore/Device/DeviceRegistrationManager.swift
```diff
@@ -0,0 +1,106 @@
...
+                    "periodic refresh"
+
+        do {
+            Logger.info("Registering device (\(reason), token: \(pushToken != nil ? "present" : "nil"))")
+
+            // Register device using AppCheck (handled by API client)
+            try await apiClient.registerDevice(deviceId: deviceId, pushToken: pushToken)
+
+            // Only persist on SUCCESS - ensures retry on failure
+            UserDefaults.standard.set(true, forKey: hasRegisteredKey)
```
jarodl: should this only happen when the push token is not `nil`?

ConvosCore/Sources/ConvosCore/Inboxes/ConversationStateMachine.swift
```diff
@@ -586,16 +616,38 @@ public actor ConversationStateMachine {
...
+        // Make optional - don't exit early if missing, only skip push notification cleanup
+        let identity = try? await identityStore.identity(for: client.inboxId)
+        if identity == nil {
+            Logger.warning("Identity not found, skipping push notification cleanup for: \(client.inboxId)")
+        }
+
+        let clientId = identity?.clientId
+
+        // Only perform push notification cleanup if we have a clientId
+        if let clientId = clientId {
```
jarodl: this might read better if it's combined with the identity == nil check above

ConvosCore/Sources/ConvosCore/Inboxes/ConversationStateMachine.swift
```diff
@@ -489,13 +507,25 @@ public actor ConversationStateMachine {
...
                     _ = try await inviteWriter.generate(for: dbConversation, expiresAt: nil)
 
-                    // Subscribe to push notifications
-                    let topic = conversation.id.xmtpGroupTopicFormat
-                    do {
-                        try await apiClient.subscribeToTopics(installationId: client.installationId, topics: [topic])
-                        Logger.info("Subscribed to push topic after join: \(topic)")
-                    } catch {
-                        Logger.error("Failed subscribing to topic after join \(topic): \(error)")
+                    // Subscribe to push notifications using clientId from keychain
```
jarodl: let's break this out into a helper method so it's not repeated

ConvosCore/Sources/ConvosCore/Inboxes/ConversationStateMachine.swift
lourou: Yes, done!

PR #200, opened 2025-10-31
Compact conversation metadata

yewreeka:
### Store conversation inbox IDs as bytes and write expiration as sfixed64 to compact conversation metadata
Switch protobuf schema and usage to `bytes` for `ConversationProfile.inboxID` and `sfixed64` seconds for `ConversationCustomMetadata.expiresAtUnix`; update conversions to validate inbox ID hex and fail on invalid or empty input; adjust invite creation to reject empty/invalid creator inbox ID and write raw bytes; update getters/setters and lookups to operate on bytes and unix timestamps.

#### 📍Where to Start
Start with the schema changes in [conversation_custom_metadata.proto](https://github.com/ephemeraHQ/convos-ios/pull/200/files#diff-aa1526b4b36cc63a5a2b823886bc97d613369263cb9cffa550ecd4853eae9d96), then review the generated code in [conversation_custom_metadata.pb.swift](https://github.com/ephemeraHQ/convos-ios/pull/200/files#diff-92297201ce3a26b985160c1bf1f910c2ad33d018acf06f3b05a5713e6222839c) and the conversion logic in [ConversationCustomMetadataExtensions.swift](https://github.com/ephemeraHQ/convos-ios/pull/200/files#diff-f6ec8bb948cb457536011eb22cc2734f19a7751684b198c15f6f11208832c563).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 24b8a62. 7 files reviewed, 15 issues evaluated, 14 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift — 0 comments posted, 10 evaluated, 10 filtered</summary>

- [line 48](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L48): External contract change: `MemberProfile.conversationProfile` was changed from non-optional (`ConversationProfile`) to optional (`ConversationProfile?`). Callers that previously relied on a non-nil value may now receive `nil` at runtime, leading to unexpected behavior (e.g., force-unwrap crashes or logic assuming a value). There is no guard or fallback in the computed property to preserve previous guarantees. <b>[ Low confidence ]</b>
- [line 49](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L49): Silent failure path: The computed property returns `ConversationProfile(inboxIdString: inboxId, name: name, imageUrl: avatar)` directly. If `ConversationProfile` uses a failable initializer for `inboxIdString` (e.g., when `inboxId` is empty or invalid), the property will evaluate to `nil` with no validation, logging, or fallback, conflating 'no profile' with 'invalid profile data'. This introduces ambiguity and could cause downstream logic to misinterpret the reason for absence. <b>[ Low confidence ]</b>
- [line 73](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L73): Potential time unit mismatch: `expiresAt` now returns `Date(timeIntervalSince1970: TimeInterval(metadata.expiresAtUnix))`. If `metadata.expiresAtUnix` is in milliseconds (common in some schemas) rather than seconds, this will produce an incorrect `Date` (far in the future), whereas the prior `metadata.expiresAt.date` may have handled unit normalization. There is no validation or conversion to ensure the correct units. <b>[ Low confidence ]</b>
- [line 79](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L79): Potential precision loss and contract change when storing expiration: `updateExpiresAt(date:)` now sets `customMetadata.expiresAtUnix = Int64(date.timeIntervalSince1970)`, truncating fractional seconds (previously `expiresAt` likely retained sub-second precision via a richer timestamp type). This can round down the intended expiry by up to 0.999s, and represents a format change in the serialized metadata field. While downstream code in this diff reads `expiresAtUnix`, any components still expecting the previous `expiresAt` field or millisecond/nanosecond precision may misinterpret or miss updates. If sub-second precision matters or backward compatibility is required, consider preserving precision (e.g., store milliseconds or nanoseconds) or populating both fields during a transition period. <b>[ Low confidence ]</b>
- [line 141](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L141): Contract parity/regression risk: switching from `inboxID` to `inboxIdString` modifies the externally visible representation of the inbox id provided to `MemberProfile`. If callers downstream expect the canonical binary/typed ID (or a specific normalized format), returning a string representation may change semantics (e.g., encoding, case, prefixes) and break equality, deduplication, or routing. Verify `MemberProfile`'s `inboxId` contract and ensure consistent normalization (e.g., always canonicalize to the same format used elsewhere) or explicitly document the change, and migrate all consumers. <b>[ Low confidence ]</b>
- [line 141](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L141): The getter constructs `MemberProfile` with `inboxId: $0.inboxIdString` without any presence/validity check, unlike `name` and `avatar` which are guarded by `hasName`/`hasImage`. If the underlying protobuf allows the inbox identifier to be absent or defaulted to an empty string, this path will produce `MemberProfile` instances with an invalid `inboxId` (e.g., empty), silently violating the output contract. At minimum, validate presence (e.g., check a `hasInboxID` flag or non-empty string) and either throw, skip such entries, or set `inboxId` to `nil` if the initializer permits optional values. <b>[ Low confidence ]</b>
- [line 191](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L191): Compression marker collision can cause hard decode failure without fallback. In `ConversationCustomMetadata.fromCompactString`, the code treats any payload whose first byte equals `Data.compressionMarker` as compressed and attempts to decompress it. If an uncompressed protobuf payload happens to begin with the same byte value (a realistic possibility unless the marker is provably impossible as a first byte for protobuf-serialized messages), `decompressedWithSize(maxSize:)` will fail its format checks and the function throws `DecodingError.dataCorrupted`. This yields an unrecoverable failure for otherwise valid metadata. To avoid false positives and ensure robustness, either use a multi-byte magic header unlikely to collide, embed a distinct container format for compressed data, or add a fallback: if decompression fails, treat the payload as uncompressed protobuf bytes rather than throwing. <b>[ Low confidence ]</b>
- [line 263](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L263): The failable initializer `ConversationProfile.init?(inboxIdString:name:imageUrl:)` accepts any non-empty hex string and does not validate the expected length (32 bytes for a 64-character XMTP inbox ID). This contradicts the documented format (“XMTP v3 format” where inbox IDs are 64-character hex strings representing 32 bytes) and can allow malformed inbox IDs to be stored, potentially breaking equality checks and downstream logic. Add a length check (e.g., `inboxIdBytes.count == 32`) to ensure only valid IDs are accepted. <b>[ Low confidence ]</b>
- [line 301](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L301): The method `ConversationCustomMetadata.removeProfile(inboxId:)` parses the hex inbox ID into `Data` but does not reject an empty hex string. `Data(hexString:)` returns an empty `Data` for an empty string, so passing `""` could remove the first profile whose `inboxID` is also empty (e.g., default-initialized), causing unintended deletions. Add an explicit `!inboxIdBytes.isEmpty` check in the guard to ensure empty inputs are rejected. <b>[ Low confidence ]</b>
- [line 315](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift#L315): The method `ConversationCustomMetadata.findProfile(inboxId:)` does not reject an empty hex string. Since `Data(hexString:)` returns an empty `Data` for empty input, passing `""` can return the first profile where `inboxID` is empty (e.g., default). This is an unintended match and inconsistent with the documented expectation of a valid hex inbox ID. Add `!inboxIdBytes.isEmpty` to the guard to return `nil` for empty inputs. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Protobuf Models/proto/conversation_custom_metadata.proto — 0 comments posted, 5 evaluated, 4 filtered</summary>

- [line 14](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/proto/conversation_custom_metadata.proto#L14): The schema allows `repeated ConversationProfile profiles` to include duplicate or out-of-order elements without a way to enforce uniqueness or adjacency constraints. If downstream logic assumes unique participants or order-sensitive behavior, this can lead to inconsistent state or duplicated effects at runtime. <b>[ Code style ]</b>
- [line 15](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/proto/conversation_custom_metadata.proto#L15): Replacing `google.protobuf.Timestamp` with `optional sfixed64 expiresAtUnix` introduces ambiguity and potential precision loss: it’s unclear whether the value is seconds or milliseconds, and any previous nanosecond precision in `Timestamp` is lost. Without an explicit unit guarantee in the schema, producers and consumers can disagree, yielding incorrect expiration times at runtime. <b>[ Low confidence ]</b>
- [line 15](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/proto/conversation_custom_metadata.proto#L15): Field `expiresAtUnix` changed wire type and semantic contract from the previous `google.protobuf.Timestamp` to `optional sfixed64` at tag `4`. This breaks backward/forward compatibility at runtime: existing encoded data with field 4 as a length-delimited `Timestamp` will be treated as an unknown/mismatched wire type by decoders expecting an 8-byte fixed integer and will be skipped or fail to parse; new data encoded as `sfixed64` will be ignored by decoders expecting a `Timestamp`. Additionally, converting from `Timestamp` to a single 64-bit value can silently lose nanosecond precision and any constraints/validation provided by the well-known type. <b>[ Low confidence ]</b>
- [line 20](https://github.com/ephemeraHQ/convos-ios/blob/24b8a628202fe5c4f86fd88cbadf4db4f7e98250/ConvosCore/Sources/ConvosCore/Protobuf Models/proto/conversation_custom_metadata.proto#L20): Field `inboxId` changed from `string` to `bytes` at tag `1`. While both use wire type 2 (length-delimited), old consumers expecting a `string` may enforce UTF-8 validity and could reject random 32-byte binary data, causing parse failures or runtime errors when decoding new messages. Conversely, new producers may emit non-UTF-8 binary data per the comment (hex-decoded 32 bytes), which violates the old `string` contract. <b>[ Low confidence ]</b>
</details>

</details>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift
+++ b/ConvosCore/Sources/ConvosCore/Protobuf Models/ConversationCustomMetadataExtensions.swift
@@ -28,6 +28,7 @@ import XMTPiOS
 enum ConversationCustomMetadataError: Error, LocalizedError {
     case randomGenerationFailed
     case invalidLength(Int)
+    case invalidInboxIdHex(String)
 
     var errorDescription: String? {
         switch self {
@@ -35,6 +36,8 @@ enum ConversationCustomMetadataError: Error, LocalizedError {
             return "Failed to generate secure random bytes"
         case .invalidLength(let length):
             return "Invalid length for random string generation: \(length). Length must be positive."
+        case .invalidInboxIdHex(let inboxId):
+            return "Failed to convert MemberProfile to ConversationProfile - invalid inbox ID hex: \(inboxId)"
         }
     }
 }
@@ -145,8 +148,7 @@ extension XMTPiOS.Group {
 
     public func updateProfile(_ profile: MemberProfile) async throws {
         guard let conversationProfile = profile.conversationProfile else {
-            Logger.warning("Failed to convert MemberProfile to ConversationProfile - invalid inbox ID hex: \(profile.inboxId)")
-            return
+            throw ConversationCustomMetadataError.invalidInboxIdHex(profile.inboxId)
         }
         var customMetadata = try currentCustomMetadata
         customMetadata.upsertProfile(conversationProfile)


yewreeka:
### Merge activity

* **Nov 1, 12:52 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/ephemeraHQ/convos-ios/200).
* **Nov 1, 12:55 AM UTC**: [Graphite](https://app.graphite.dev/github/pr/ephemeraHQ/convos-ios/200) rebased this pull request as part of a merge.
* **Nov 1, 12:56 AM UTC**: @yewreeka merged this pull request with [Graphite](https://app.graphite.dev/github/pr/ephemeraHQ/convos-ios/200).

PR #230, opened 2025-11-08
Cache images to disk

yewreeka:
### Add disk-backed image caching and modify UI and writers to use async ImageCache reads in [ImageCache.swift](https://github.com/ephemeraHQ/convos-ios/pull/230/files#diff-9c43db5e1255cd76ef671f19b4a4c7a02209590452cefbe49fbee60b9884b12b)
Introduce disk persistence, async retrieval, and LRU cleanup in `ImageCache`; switch avatar and notification flows to `imageAsync`; key conversation images by `clientConversationId`; add single-pass JPEG/PNG compression utilities and cache QR codes as PNG.

#### 📍Where to Start
Start with the `ImageCache` initialization and async retrieval paths in [ImageCache.swift](https://github.com/ephemeraHQ/convos-ios/pull/230/files#diff-9c43db5e1255cd76ef671f19b4a4c7a02209590452cefbe49fbee60b9884b12b).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized fdaef0d. 12 files reviewed, 37 issues evaluated, 35 issues filtered, 1 comment posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>Convos/Debug View/DebugView.swift — 0 comments posted, 3 evaluated, 3 filtered</summary>

- [line 257](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/Convos/Debug View/DebugView.swift#L257): `testSentryMessage()` was changed from `private` to internal, making it callable from outside the extension’s file/module scope. In production builds or from unintended callers, this can emit Sentry events when invoked, creating noisy telemetry and side effects. There is no guard against environment or Sentry initialization state here. <b>[ Low confidence ]</b>
- [line 263](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/Convos/Debug View/DebugView.swift#L263): `testSentryError()` was changed from `private` to internal, exposing a method that emits an error event to Sentry. When invoked outside intended debug contexts, this can generate misleading telemetry and side effects. No environment checks are present. <b>[ Low confidence ]</b>
- [line 277](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/Convos/Debug View/DebugView.swift#L277): `testSentryException()` was changed from `private` to internal, exposing a method that emits an exception event to Sentry. If invoked in production or by unintended callers, it can produce misleading telemetry and side effects. No environment gating is implemented. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Shared Views/AvatarView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 16](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/Convos/Shared Views/AvatarView.swift#L16): Using `placeholderImage ?? cachedImage` in `AvatarView` means that when `placeholderImage` is non-nil, it will always be chosen and the loaded/cached image in `cachedImage` will never be displayed. This silently prevents transitioning from a placeholder to the real image, causing the avatar to remain stuck on the placeholder even after successful network/cache load. <b>[ Low confidence ]</b>
- [line 101](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/Convos/Shared Views/AvatarView.swift#L101): In `ConversationAvatarView`, `fallbackName` is set to an empty string (`""`). When neither `conversation.imageURL` nor `conversationImage` is available (the else branch uses `MonogramView(text: "")`), and even when the `AvatarView` shows the monogram, this change results in an empty monogram instead of using `conversation.displayName`. This is an externally visible regression in fallback behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Extensions/ImageCacheableExtensions.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 11](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Extensions/ImageCacheableExtensions.swift#L11): Changing `Conversation.imageCacheIdentifier` from `id` to `clientConversationId` alters the cache key used across memory and disk caches. Any previously cached images keyed by `id` will no longer be retrievable, leading to silent cache misses and orphaned disk entries. If `clientConversationId` is not globally unique or can change over time, this also risks key collisions and stale cache behavior. This is an externally visible contract change for `ImageCacheable` that affects `imageAsync(for:)`, which now looks up using the new key. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift — 0 comments posted, 7 evaluated, 6 filtered</summary>

- [line 67](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift#L67): `maxDiskCacheSize` is taken directly from `CacheConfiguration.maxDiskCacheSize` without validation. If this value is zero or negative, cleanup will run and potentially remove all cached files even when the cache may be intentionally disabled or misconfigured. Add a lower-bound guard (e.g., treat non-positive as no-op or clamp to minimum) to avoid aggressive deletion due to bad config. <b>[ Low confidence ]</b>
- [line 201](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift#L201): Misleading parameter `imageFormat` in `image(for identifier: String, imageFormat: ImageFormat = .jpg)` is unused. The function is memory-only and ignores `imageFormat`, which suggests format influences retrieval. This contradicts the visible signature/contract and can cause confusion or incorrect usage paths expecting format-sensitive behavior. <b>[ Low confidence ]</b>
- [line 214](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift#L214): `image(for identifier: String, imageFormat: ImageFormat = .jpg) -> UIImage?` ignores its `imageFormat` parameter and always returns the memory-cached image for the identifier regardless of format. This creates contract inconsistency with `imageAsync(for:imageFormat:)`, which honors the format for disk reads. If the same `identifier` is used to cache different formats at different times (e.g., PNG with transparency vs. JPEG), callers may request a specific format but receive whichever image happens to be in memory, leading to incorrect rendering (e.g., lost alpha). Consider either removing the `imageFormat` parameter from the synchronous method or ensuring memory-tier caches are segregated by format or validated against the requested format. <b>[ Low confidence ]</b>
- [line 307](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift#L307): API asymmetry for URL-based caching can cause cache misses. `image(for url: URL)` and `imageAsync(for url: URL)` use `url.absoluteString` as the key, but the setter is `setImage(_ image: UIImage, for url: String)` which requires callers to pass an exact matching string. Different string representations of the same `URL` (e.g., differing percent-encoding, trailing slashes, or relative vs. absolute) will lead to keys that don’t match, resulting in missed reads. Provide a `setImage(_:for url: URL)` overload that uses the same normalization (`url.absoluteString`) or standardize on one type. <b>[ Low confidence ]</b>
- [line 546](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift#L546): `removedSize` accounting can be incorrect when entries represent directories or when `fileSize` is unavailable (`resourceValues.fileSize` is `nil`), because size defaults to `0`. Deleting such a URL may remove non-trivial data while `removedSize` stays low, causing inaccurate logs and possibly continuing deletions longer than necessary. Consider skipping directories, computing actual sizes, or updating `totalSize` by re-stat after deletions. <b>[ Low confidence ]</b>
- [line 553](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift#L553): The implementation explicitly contradicts its comments about memory usage. Despite stating "Processes files in batches to avoid memory pressure", it accumulates all `CachedFileInfo` entries into `oldestFiles` and sorts the entire collection (`oldestFiles.append(contentsOf: batch)` then `oldestFiles.sort`). For very large caches, this can cause high memory usage and potential OOM. Process incrementally: either select candidates with a bounded heap, or delete while iterating without storing all entries. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 217](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift#L217): Temporary file created for the `UNNotificationAttachment` is never deleted, causing a disk leak. The function writes the cached image to `tempFileURL` (`Data.write(to:)` at line `217`) and uses it to create an attachment, but there is no corresponding cleanup on success or failure paths. If attachment creation or request scheduling succeeds or fails, the file remains in the temporary directory. Add explicit removal of `tempFileURL` after the notification is scheduled (or after attachment creation when safe), and ensure cleanup occurs on all paths, including the `catch` block. A typical fix is to call `try? FileManager.default.removeItem(at: tempFileURL)` in a `defer` or after `UNUserNotificationCenter.current().add(request)` completes successfully. <b>[ Out of scope ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationsRepository.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 25](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationsRepository.swift#L25): Parameter `date` in `Conversation.mock(...)` is ignored. The initializer sets `createdAt` to `Date()` instead of the passed `date`, so callers cannot control the conversation's creation timestamp. Replace `createdAt: Date()` with `createdAt: date` to honor the parameter. <b>[ Out of scope ]</b>
- [line 51](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Storage/Repositories/Mocks/MockConversationsRepository.swift#L51): `Conversation.empty(...)` sets `createdAt` to `.distantFuture`. Any relative-time logic that computes `referenceDate.timeIntervalSince(createdAt)` (e.g., `Date.relativeShort(to:)`) will yield negative values, producing outputs like "-1h" or "-2w" and potentially breaking code that assumes non-future creation dates. Use a past or present timestamp (e.g., `.distantPast` or `.init()`) for drafts, or explicitly handle future dates in consumers. <b>[ Out of scope ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationMetadataWriter.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 158](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationMetadataWriter.swift#L158): The shift to asynchronous caching via `ImageCache.shared.resizeCacheAndGetData(_, for:)` in both `updateImage(_:, for:)` (lines `158`–`165`) and `update(avatar:conversationId:)` (lines `91`–`96`) alters contract timing compared to prior synchronous `setImage` usage. Callers may immediately attempt to read from the cache after invocation and observe a miss until the `Task` completes, causing transient UI inconsistencies. Additionally, `Conversation.imageCacheIdentifier` was changed to `clientConversationId` (from `id`) elsewhere, which can orphan previously cached entries keyed by `id`. Consider documenting the timing change, ensuring consumers rely on `cacheUpdateSubject` before reading, and migrating/cleaning old cache keys to avoid stale disk entries. <b>[ Low confidence ]</b>
- [line 171](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Storage/Writers/ConversationMetadataWriter.swift#L171): In `updateImage(_:, for:)` (lines `167`–`176`), errors thrown inside the closure passed to `uploadAttachmentAndExecute` are caught and only logged (`Log.error`), without propagation or compensating actions. If the upload succeeds but `updateImageUrl` fails, the system may end up with an uploaded image whose URL is not reflected in the conversation metadata or local database, leaving inconsistent state. Consider propagating the error or implementing rollback/retry to maintain consistency. <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Storage/Writers/MyProfileWriter.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 85](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Storage/Writers/MyProfileWriter.swift#L85): In `update(avatar:conversationId:)`, when `avatar` is `nil` (lines `82`–`88`), the code removes the cached image and updates the remote profile via `group.updateProfile(updatedProfile)` but returns without persisting the `avatar: nil` change to the local database. This creates a local/remote state inconsistency where the remote profile has no avatar but the local `MemberProfile` may still hold the old `avatar` URL. You should save the updated profile to the database in the `nil` branch (mirroring the non-nil branch) before returning. <b>[ Already posted ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift — 0 comments posted, 9 evaluated, 9 filtered</summary>

- [line 22](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L22): `UIImage` instances backed only by `CIImage` will have `image.cgImage == nil`, causing the function to return `nil` for otherwise valid inputs. This is a reachable case (e.g., images created via Core Image). To avoid silent failure, render the `CIImage` to a `CGImage` using a `CIContext` fallback when `cgImage` is nil. <b>[ Low confidence ]</b>
- [line 60](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L60): `CGContext` is created with `width: Int(targetPixelSize.width)` and `height: Int(targetPixelSize.height)`. If either target dimension is between 0 and 1 pixel (due to small `maxSize`, large downscale, or fractional computation), truncation to `Int` yields 0, causing `CGContext` creation to fail and the function to return `nil`. Clamp to at least 1 pixel and round (e.g., `max(1, Int(round(...)))`). <b>[ Already posted ]</b>
- [line 60](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L60): `maxSize` is not validated for non-positive values, and `targetPixelSize` can become zero or negative. This flows into `CGContext` creation with `width`/`height` as `Int(targetPixelSize.width/height)` (lines `60`-`61`), which will fail and return `nil`. While this avoids a crash, it yields unexpected failure for reachable inputs (e.g., `maxSize` of `0` or negative). Enforce minimum dimensions (>= 1) and non-negative max size before creating the context, or clamp the computed `targetPixelSize` to at least 1×1 pixels. The same issue exists in `resizeAndCompressToPNG`. <b>[ Already posted ]</b>
- [line 60](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L60): Rounding/truncation when converting `targetPixelSize` from `CGFloat` to `Int` can introduce off-by-one distortion or unintended aspect bias. Using `Int` truncation may slightly shrink one or both dimensions, degrading output quality. Prefer rounding to nearest (`Int(round(...))`) and ensuring consistent aspect by computing the second dimension from the first via the original ratio after rounding. <b>[ Already posted ]</b>
- [line 108](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L108): `compressionQuality` is documented as 0.0–1.0 but is not validated. Passing values outside this range (including NaN) can lead to undefined or driver-specific behavior in ImageIO. Add a guard to clamp or reject out-of-range values before constructing the destination options. <b>[ Low confidence ]</b>
- [line 109](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L109): `compressionQuality` in `resizeAndCompressToJPEG` is not validated. Values outside `[0.0, 1.0]` are reachable and can lead to undefined or default behavior in ImageIO (`kCGImageDestinationLossyCompressionQuality` at line `109`). Clamp or validate the quality to the supported range to ensure predictable output. <b>[ Low confidence ]</b>
- [line 261](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L261): `checkForTransparentPixels(cgImage:)` samples every `sampleStep` pixels instead of scanning all pixels. This can return `false` (no transparency) when transparent pixels exist but fall between the sampling strides, especially on larger images where `sampleStep` grows (line 261). This is a logical false-negative that violates the intended semantics of `hasTransparency(_:)`, which claims to detect whether an image "has transparency". <b>[ Low confidence ]</b>
- [line 276](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L276): `checkForTransparentPixels(cgImage:)` returns `false` when the `CGContext` cannot be created (line 276) or when `context.data` is `nil` (line 285). These are operational failures (e.g., memory pressure) but they are treated as "no transparency" rather than error states, causing silent false negatives and violating caller expectations that `false` strictly means "no transparent pixels present". <b>[ Low confidence ]</b>
- [line 320](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/ImageCompression.swift#L320): Potential off-by-one/cropping due to fractional pixel sizes: `targetSize` (derived from `targetPixelSize`) can contain fractional values, but the `CGContext` is created with `Int(targetPixelSize.width/height)` (truncating). The transforms and `drawingRect` here use `targetSize.width`/`height` directly, which may be larger than the actual bitmap by up to almost 1 pixel. This can cause subtle clipping or empty borders on some orientations/scales. To avoid this, round `targetPixelSize` to integers first and use those exact integer dimensions consistently for both the context creation and for all transform/drawing calculations (including the swapped sizes in 90° rotations). <b>[ Low confidence ]</b>
</details>

<details>
<summary>ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift — 1 comment posted, 9 evaluated, 8 filtered</summary>

- [line 98](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L98): The `scale` parameter is effectively ignored throughout. `OptionsHashKey` hard-sets `scale` to `3.0`, the cache key uses normalized scale, and rendering uses `normalizedScale` set to `3.0` when creating the `UIImage`. As a result, passing a custom `scale` (including via `pregenerate(from:scale:)`) does not affect cache keys or output image DPI. This is a contract mismatch and can lead to unexpected caching collisions or incorrect expectations about image scale. <b>[ Low confidence ]</b>
- [line 178](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L178): `centerSpaceSize` is documented as 0.0–1.0 but there is no validation; out-of-range values are passed to `filter.centerSpaceSize`, potentially producing invalid or failed output (`outputImage == nil`). Clamp or validate the value. <b>[ Out of scope ]</b>
- [line 179](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L179): `correctionLevel` accepts arbitrary strings and is passed directly to `filter.correctionLevel`. Invalid values (not one of "L","M","Q","H") can cause `outputImage` to be `nil`, silently failing. Validate and normalize (uppercase) the input or provide a safe default. <b>[ Out of scope ]</b>
- [line 188](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L188): `Options.scale` is documented as "The scale factor to use for rendering" but `generate(from:options:)` now hard-codes `normalizedScale = 3.0` and uses it for both `targetPixelSize` and the resulting `UIImage`'s `scale`, ignoring the caller-provided `options.scale`. This is an externally visible behavior change and contradicts the documented contract, making passed `scale` values ineffective. <b>[ Low confidence ]</b>
- [line 192](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L192): `baseSize` can be zero, leading to `scaleFactor = targetPixelSize / baseSize` dividing by zero and yielding `infinite` or `nan`. This makes `createCGImage` likely fail and the function returns `nil` unpredictably. Guard `baseSize > 0` before computing the scale, or early-return with a clear failure. Involved code: `let baseSize = max(outputExtent.width, outputExtent.height)` and `let scaleFactor = targetPixelSize / baseSize`. <b>[ Out of scope ]</b>
- [line 221](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L221): `generate(from:options:) async` uses `Task { ... }` rather than `Task.detached`. This inherits the current actor and can execute heavy Core Image work on the main actor if called from UI, causing UI jank or deadlocks. Use `Task.detached(priority:)` or ensure rendering executes off the main actor (e.g., with `@Sendable` work on a background executor). <b>[ Low confidence ]</b>
- [line 258](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L258): `clearFromCache(string:options:)` calls `ImageCache.shared.removeImage(for:)` without specifying an `imageFormat`, while generation and retrieval now consistently use `.png`. If `removeImage(for:)` only clears memory or a default format, the PNG on disk may remain and be reloaded by `imageAsync(for:imageFormat:)`, causing cache entries to persist unexpectedly after a clear. <b>[ Low confidence ]</b>
- [line 263](https://github.com/ephemeraHQ/convos-ios/blob/fdaef0d69638a6d25bce0086ba4165334e0e2ab0/ConvosCore/Sources/ConvosCore/Utilities/QRCodeGenerator.swift#L263): `clearFromCacheAllModes(string:)` uses `Options.qrCodeLight`/`qrCodeDark` which are defined with `backgroundColor: .clear`, while `pregenerate(from:)` uses `backgroundColor: .white` and `.black`. This mismatch means `clearFromCacheAllModes` may not clear images generated by `pregenerate`, violating caller expectations based on the function’s documentation and name. <b>[ Out of scope ]</b>
</details>

</details>

### DIFF
--- a/Convos/Debug View/DebugView.swift
+++ b/Convos/Debug View/DebugView.swift
@@ -254,13 +254,13 @@ extension DebugViewSection {
         }
     }
 
-    private func testSentryMessage() {
+    func testSentryMessage() {
         let message = "Test message from local development - \(Date())"
         SentrySDK.capture(message: message)
         Log.info("Sent Sentry test message: \(message)")
     }
 
-    private func testSentryError() {
+    func testSentryError() {
         let error = NSError(
             domain: "com.convos.debug",
             code: 999,
@@ -274,7 +274,7 @@ extension DebugViewSection {
         Log.info("Sent Sentry test error")
     }
 
-    private func testSentryException() {
+    func testSentryException() {
         let exception = NSException(
             name: .init("TestException"),
             reason: "Test exception from local debug view",
@@ -287,7 +287,7 @@ extension DebugViewSection {
         Log.info("Sent Sentry test exception")
     }
 
-    private func testSentryWithBreadcrumbs() {
+    func testSentryWithBreadcrumbs() {
         let crumb1 = Breadcrumb(level: .info, category: "navigation")
         crumb1.message = "User navigated to Debug view"
         crumb1.data = ["screen": "DebugView"]
--- a/Convos/Shared Views/AvatarView.swift
+++ b/Convos/Shared Views/AvatarView.swift
@@ -36,9 +36,9 @@ struct AvatarView: View {
     @MainActor
     private func loadImage() async {
         // First check object cache for instant updates
-        if let cachedObjectImage = ImageCache.shared.image(for: cacheableObject) {
+        if let cachedObjectImage = await ImageCache.shared.imageAsync(for: cacheableObject) {
             cachedImage = cachedObjectImage
-            return
+            return // Early return to avoid redundant URL cache check
         }
 
         guard let imageURL else {
@@ -46,9 +46,10 @@ struct AvatarView: View {
             return
         }
 
-        // Check URL-based cache
-        if let existingImage = ImageCache.shared.image(for: imageURL) {
+        // Check URL-based cache (only if object cache was empty)
+        if let existingImage = await ImageCache.shared.imageAsync(for: imageURL) {
             cachedImage = existingImage
+            return // Early return to avoid network request if URL cache has the image
         }
 
         isLoading = true
@@ -97,7 +98,7 @@ struct ConversationAvatarView: View {
             // Fall back to URL-based loading with conversation object for cache awareness
             AvatarView(
                 imageURL: conversation.imageURL,
-                fallbackName: conversation.displayName,
+                fallbackName: "",
                 cacheableObject: conversation,
                 placeholderImage: conversationImage
             )
--- a/ConvosCore/Sources/ConvosCore/Extensions/ImageCacheableExtensions.swift
+++ b/ConvosCore/Sources/ConvosCore/Extensions/ImageCacheableExtensions.swift
@@ -8,6 +8,6 @@ extension Profile: ImageCacheable {
 
 extension Conversation: ImageCacheable {
     public var imageCacheIdentifier: String {
-        id
+        clientConversationId
     }
 }
--- a/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift
+++ b/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift
@@ -1,6 +1,24 @@
 import Combine
+import CryptoKit
+import Foundation
 import Observation
+import os
 import SwiftUI
+
+// MARK: - Image Format
+
+/// Image format for caching
+public enum ImageFormat: Sendable {
+    case jpg
+    case png
+
+    var fileExtension: String {
+        switch self {
+        case .jpg: return ".jpg"
+        case .png: return ".png"
+        }
+    }
+}
 
 // MARK: - ImageCacheable Protocol
 
@@ -10,16 +28,46 @@ public protocol ImageCacheable {
     var imageCacheIdentifier: String { get }
 }
 
+// MARK: - Cache Configuration
+
+/// Configuration for image cache limits and sizes
+private struct CacheConfiguration {
+    /// Maximum size of disk cache in bytes (500MB)
+    static let maxDiskCacheSize: Int = 500 * 1024 * 1024
+
+    /// Maximum number of images in the main memory cache
+    static let memoryCacheCountLimit: Int = 400
+
+    /// Maximum total cost (in bytes) for the main memory cache (200MB)
+    static let memoryCacheTotalCostLimit: Int = 200 * 1024 * 1024
+
+    /// Maximum number of images in the URL memory cache
+    static let urlCacheCountLimit: Int = 200
+
+    /// Maximum total cost (in bytes) for the URL memory cache (100MB)
+    static let urlCacheTotalCostLimit: Int = 100 * 1024 * 1024
+}
+
 // MARK: - Generic Image Cache
 
 /// Smart reactive image cache that stores images for any ImageCacheable object with instant updates.
+/// Supports three-tier caching: memory → disk → network
 /// When a new image is uploaded for an object, all views showing that object update instantly.
 @Observable
-public final class ImageCache {
+public final class ImageCache: @unchecked Sendable {
     public static let shared: ImageCache = ImageCache()
 
     private let cache: NSCache<NSString, UIImage>
     private let urlCache: NSCache<NSString, UIImage>
+
+    // Disk cache properties
+    private let diskCacheURL: URL
+    private let diskCacheQueue: DispatchQueue
+    private let fileManager: FileManager
+    private let maxDiskCacheSize: Int = CacheConfiguration.maxDiskCacheSize
+
+    // Cleanup coordination
+    private let cleanupLock: OSAllocatedUnfairLock<Task<Void, Never>?> = OSAllocatedUnfairLock(initialState: nil)
 
     /// Publisher for specific cache updates by identifier
     private let cacheUpdateSubject: PassthroughSubject<String, Never> = PassthroughSubject<String, Never>()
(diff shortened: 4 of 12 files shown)


yewreeka:
* **#230** <a href="https://app.graphite.com/github/pr/ephemeraHQ/convos-ios/230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ephemeraHQ/convos-ios/230?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ephemeraHQ?ref=merge-queue-instructions-in-stack-comment&prId=230).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #238, opened 2025-11-13
Add max values for dynamic type, use styles instead of points

yewreeka:
### Constrain Dynamic Type scaling across settings, conversations, and profiles to specific max sizes and replace fixed point fonts with semantic styles in SwiftUI views
Switch multiple labels and icons to semantic SwiftUI text styles and apply `View.dynamicTypeSize(...)` caps (e.g., `xxxLarge`, `accessibility1`, `accessibility2`) to key screens and containers. Key typography updates include `callout`, `body`, `footnote`, `caption`, and `caption2` replacing fixed point sizes in list items, toolbars, input bars, and overlays.

#### 📍Where to Start
Start with the Dynamic Type caps in `ConversationView` root at [ConversationView.swift](https://github.com/ephemeraHQ/convos-ios/pull/238/files#diff-8d85226aa6385ce7a0aca11596c32eb20e3c3e89b37f51651c35fceb5e28f6bc), then review shared presenters at [ConversationInfoPresenter.swift](https://github.com/ephemeraHQ/convos-ios/pull/238/files#diff-13675ac75cb89632c88080581a5e13bc7d307cf9d0894e754bf34a5c82e8178b) and app-wide settings at [AppSettingsView.swift](https://github.com/ephemeraHQ/convos-ios/pull/238/files#diff-30e98d39613710cf0f912b477ed30a8638995959caa9cb2e9facdc10e1c3f8b9).

----

<a href="https://app.macroscope.com">Macroscope</a> summarized 4bbe040.

PR #239, opened 2025-11-14
WIP improve quickname

yewreeka:
### Introduce `FocusCoordinator` across Conversation screens to coordinate quickname-related focus transitions and cap Dynamic Type scaling to xxxLarge/accessibility1 in views such as `ConversationView`, `ConversationInfoPresenter`, and `MessagesBottomBar`
Replace ad‑hoc focus handling with a centralized `FocusCoordinator`, wiring it through Conversation views and presenters, and update typography to semantic text styles while capping Dynamic Type in key screens.

#### 📍Where to Start
Start with the `FocusCoordinator` implementation and its APIs in [FocusCoordinator.swift](https://github.com/ephemeraHQ/convos-ios/pull/239/files#diff-40b3d28b766ebc0135d2e62700a55d912572d14966c5a3b6f048955c7ca841d7), then follow its integration in `ConversationViewModel` in [ConversationViewModel.swift](https://github.com/ephemeraHQ/convos-ios/pull/239/files#diff-d056b0b131534648b6d13cddafd053be956f5d23d035ad2538db85a8cc370d8d) and the presenter in [ConversationInfoPresenter.swift](https://github.com/ephemeraHQ/convos-ios/pull/239/files#diff-13675ac75cb89632c88080581a5e13bc7d307cf9d0894e754bf34a5c82e8178b).

----

<details>
<summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized e30fd3c. 29 files reviewed, 48 issues evaluated, 42 issues filtered, 0 comments posted</summary>

### 🗂️ Filtered Issues
<details>
<summary>Convos/App Settings/AppSettingsView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 112](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/App Settings/AppSettingsView.swift#L112): `Button` wraps a `NavigationLink` inside its label (lines 109–126). Nesting interactive controls can cause unpredictable runtime behavior: taps may both trigger `openURL(...)` and attempt to navigate to `EmptyView`, or the inner `NavigationLink` may be inert while still creating an accessibility focus target. This violates "at-most-once" action semantics and can yield inconsistent outcomes across platforms/OS versions. Use a non-interactive view (e.g., `HStack`) styled as a row, or make the outer element a `NavigationLink` with a `Link`/`openURL` action, but not both. <b>[ Low confidence ]</b>
- [line 140](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/App Settings/AppSettingsView.swift#L140): A second instance of a `Button` wrapping a `NavigationLink` exists for "Privacy & Terms" (lines 137–142). This duplicates the interactive nesting bug: tapping may both trigger `openURL(...)` and attempt to navigate to `EmptyView`, or create conflicting accessibility focus targets. Replace the inner `NavigationLink` with a non-interactive view, or use a single control for the intended action. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Creation/JoinConversationView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 87](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Creation/JoinConversationView.swift#L87): Force-unwrap of URL literal in `openURL(URL(string: "https://convos.org/terms-and-privacy")!)` can crash at runtime if the string ever becomes malformed (e.g., future edit/localization). Prefer safe construction (`if let url = URL(string: ...) { openURL(url) }`), or store a prevalidated URL constant. <b>[ Out of scope ]</b>
- [line 143](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Creation/JoinConversationView.swift#L143): `onScannedCode` can be invoked multiple times without any guarding or debouncing: it’s called from `.onChange(of: viewModel.scannedCode)` whenever `scannedCode` changes, and also from the clipboard button. If the scanner emits the same or successive codes rapidly (common for camera scanners) or the user taps the clipboard button repeatedly, this can trigger duplicate navigation/network actions or state mutations. Consider debouncing, single-consume semantics (swap-and-clear), or idempotency guards. <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/Conversation Creation/NewConversationView.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 71](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Creation/NewConversationView.swift#L71): The "Keep" option in the deletion confirmation (`Button("Keep")`) dismisses the view by calling `dismiss()` instead of canceling the close and keeping the conversation creation view open. This contradicts the expected behavior for a "Keep" action in a delete-confirmation flow, which should avoid deletion and not dismiss the current screen. As written, users who choose to keep the conversation will still exit the flow, potentially causing unintended navigation and loss of context. Consider removing the `dismiss()` call from the "Keep" button so it simply closes the dialog without dismissing the view. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Conversation Detail Drawer/AddQuicknameView.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 18](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Conversation Detail Drawer/AddQuicknameView.swift#L18): The `Button` action does not guard against rapid repeated taps, so `onUseProfile(profile, profileImage)` can be invoked multiple times before the overlay dismisses. This can cause double-application of side effects (e.g., navigation, state changes). Consider debouncing, disabling the button after first tap, or making `animateOut()`/`onUseProfile` idempotent. <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Conversation Detail Drawer/InviteAcceptedView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 30](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Conversation Detail Drawer/InviteAcceptedView.swift#L30): Duplicate scheduling: `.onAppear` can be invoked multiple times for the same view identity, but the code schedules a new `asyncAfter` each time without deduplication. This can create multiple pending closures all attempting `self.showingDescription = true`. While idempotent, it wastes work and can produce repeated animations or redundant state writes. Guard with a flag (e.g., only schedule if `showingDescription == false` and no pending task), or coalesce via a single cancellable `Task` stored in `@State`/`@StateObject`. <b>[ Low confidence ]</b>
- [line 31](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Conversation Detail Drawer/InviteAcceptedView.swift#L31): Lifecycle/cancellation bug: the `DispatchQueue.main.asyncAfter` scheduled in `.onAppear` (lines 31–36) is never cancelled if the view disappears before the delay elapses. The escaping closure captures `self` and will attempt `self.showingDescription = true` even if the view has been removed, which can cause state updates on a non-visible/defunct view and runtime warnings. Use a cancellable task (`.task` with `Task.sleep` and cancellation in `.onDisappear`) or store a `DispatchWorkItem` and cancel it in `.onDisappear`, and guard against updating after teardown. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Conversation Detail Drawer/UseAsQuicknameView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 18](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Conversation Detail Drawer/UseAsQuicknameView.swift#L18): The `"Use as Quickname"` button invokes `onUseAsQuickname()` with no guard against rapid re-taps. Users can tap multiple times before the overlay finishes animating out, causing `onUseAsQuickname` to be called multiple times (double-application). Add an at-most-once guard (e.g., a `@State` flag to disable the button after first tap or swap-and-clear the callback) to prevent repeated invocation. <b>[ Low confidence ]</b>
- [line 44](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Conversation Detail Drawer/UseAsQuicknameView.swift#L44): Tapping the `"Learn more about Quickname"` button calls `animateOut()` (dismissing `AnimatedOverlayView`) before setting `presentingLearnMore = true` to present the `.selfSizingSheet`. Because the sheet is attached to the overlay’s content (`VStack`), dismissing the overlay tears down the view hierarchy and the sheet cannot appear or will immediately disappear. Users cannot reliably open `WhatIsQuicknameView`. Remove the `animateOut()` call for the learn-more path, or move the sheet presentation outside the overlay so it survives dismissal. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/ConversationInfoView.swift — 0 comments posted, 2 evaluated, 1 filtered</summary>

- [line 41](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationInfoView.swift#L41): The optional subtitle is displayed whenever it is non-nil using `if let subtitle { Text(subtitle) ... }`. If `subtitle` is an empty string, the view will render a blank `Text` line and reserve vertical space, creating unintended extra spacing with no content. Consider filtering out empty strings or applying a conditional `!subtitle.isEmpty` check. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/ConversationShareView.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 33](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationShareView.swift#L33): `BackgroundClearView()` is used (line 33) to clear the presenting container’s background by setting `view.superview?.superview?.backgroundColor = .clear` but there is no corresponding restoration on dismissal. This is a persistent mutation of a shared UIKit container that can leak beyond the lifetime of this sheet, causing subsequent modals/sheets to render with a transparent background. The temporary styling must be scoped and restored on every exit path (e.g., capture and restore the prior `backgroundColor` in `updateUIView`/`dismantleUIView`, or use a non-mutating approach). <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/ConversationViewModel.swift — 0 comments posted, 7 evaluated, 7 filtered</summary>

- [line 261](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L261): Several methods now depend on an optional `FocusCoordinator` parameter and silently no-op when `nil`, which is a contract change from prior behavior where internal focus was updated directly. If callers pass `nil` or fail to forward a coordinator, focus transitions (e.g., moving to `.conversationName`, dismissing to `nil`, re-focusing message) will not occur, degrading UX with no visible fallback. This affects `onConversationInfoTap`, `onConversationNameEndedEditing`, `onConversationSettings`, `onConversationSettingsDismissed`, `onProfilePhotoTap`, `onProfileSettingsDismissed`, `onSendMessage`, and `onDisplayNameEndedEditing`. Consider making the coordinator non-optional or providing an internal fallback to preserve focus behavior when `nil`. <b>[ Low confidence ]</b>
- [line 270](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L270): Actor isolation is violated by launching non-MainActor `Task { ... }` blocks that access `self` (a `@MainActor` class) and its properties before hopping back to the main actor. This can cause undefined behavior and runtime data races under Swift concurrency. Examples: in `onConversationNameEndedEditing(...)` the `Task` at line 270 reads `self.conversation.id` and uses `metadataWriter`; similarly the `Task` at lines 286–296 and 303–313; in `onSendMessage(...)` the `Task` at line 352 reads `self.outgoingMessageWriter`; in `remove(member:)` the `Task` at line 401 reads `self.conversation.id`; in `leaveConvo()` the `Task` at line 412 reads `self.conversation.clientId`; and in `explodeConvo()` the `Task` at line 429 reads `self.conversation.members` and `self.conversation.id`. Use `Task { @MainActor in ... }` when touching `self` or capture needed values on the main actor and pass them into a background task. <b>[ Out of scope ]</b>
- [line 330](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L330): Cancellation doesn’t revert pending image changes: `onConversationSettingsCancelled()` (lines 330–335) resets name/description edit state and values but leaves `conversationImage` intact. Later `onConversationNameEndedEditing(...)` will still update the image (lines 283–297), potentially applying an image the user intended to cancel. Clear or revert `conversationImage` on cancel to avoid unintended updates. <b>[ Low confidence ]</b>
- [line 349](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L349): `onSendMessage(focusCoordinator:)` clears `messageText` before attempting to send. On failure, the message is lost with only a log entry (`Log.error`), providing no recovery or user feedback and leaving the input field empty. This is silent data loss at runtime. Consider deferring clearing until after a successful send, or restoring `messageText` on error and surfacing failure to the user. <b>[ Low confidence ]</b>
- [line 349](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L349): User input loss on send failure: `onSendMessage(...)` clears `messageText` before sending (`messageText = ""` at line 349). If `outgoingMessageWriter.send(...)` throws, the original text is lost and cannot be retried, yielding poor UX and silent data loss. Defer clearing until after a successful send or store the text for retry. <b>[ Low confidence ]</b>
- [line 385](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L385): False-positive onboarding trigger: `onDisplayNameEndedEditing(...)` computes `didChangeProfile = !profile.displayName.isEmpty || profileImage != nil` (lines 385–388) and calls `onboardingCoordinator.didChangeProfile(profile:)` whenever either is non-empty/non-nil, even if the user made no changes. This can incorrectly prompt onboarding UI on mere blur events. Compute actual change by comparing to the latest stored profile/image or reuse the change detection from `MyProfileViewModel.onEndedEditing(...)`. <b>[ Low confidence ]</b>
- [line 385](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/ConversationViewModel.swift#L385): `onDisplayNameEndedEditing(focusCoordinator:)` triggers onboarding via `didChangeProfile(profile:)` whenever `!profile.displayName.isEmpty || profileImage != nil`, which does not confirm an actual change relative to prior profile state. This can cause spurious onboarding prompts even when the user didn’t modify their profile (e.g., an existing non-empty display name). Consider comparing the edited values against the latest saved profile or using the update result from `myProfileViewModel.onEndedEditing` to determine if a change occurred. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Messages/ConversationInfoButton.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 93](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/ConversationInfoButton.swift#L93): The `.onChange(of: focusCoordinator?.currentFocus)` will not fire when `currentFocus` changes inside `FocusCoordinator` unless the view is re-rendered by some other state change, because `focusCoordinator` is a plain `let` reference and is not observed (e.g., via `@ObservedObject`/`@StateObject` or an explicit publisher). As a result, `progress` may never update in response to focus changes (e.g., when calling `focusCoordinator?.moveFocus(...)`). To reliably react to `currentFocus` updates, make `focusCoordinator` an `@ObservedObject` (non-optional or wrapped via a proxy), or use `.onReceive(focusCoordinator?.objectWillChange)`/a dedicated `Publisher` for `currentFocus`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Messages/Messages View Controller/Keyboard/KeyboardInfo.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 20](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/Messages View Controller/Keyboard/KeyboardInfo.swift#L20): Overly strict guarding on `UIResponder.keyboardIsLocalUserInfoKey`: Treating `keyboardIsLocal` as required causes the initializer to return `nil` when the key is absent. Some keyboard notifications may legitimately omit this key, leading to dropped events (e.g., `keyboardDidChangeFrame`). Provide a default (e.g., `false`) when the key is missing instead of failing the whole parse. <b>[ Low confidence ]</b>
- [line 34](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/Messages View Controller/Keyboard/KeyboardInfo.swift#L34): Unsafe enum initialization: `NSNumber(value: keyboardAnimationCurve).getValue(&animationCurve)` writes raw bytes into a `UIView.AnimationCurve` variable without validating that the integer corresponds to a valid case. This can produce an invalid enum value and undefined behavior when used later. Prefer `UIView.AnimationCurve(rawValue: keyboardAnimationCurve) ?? .easeInOut` for safe conversion with a defined fallback. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Messages/Messages View Controller/View Controller/MessagesViewController.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 549](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/Messages View Controller/View Controller/MessagesViewController.swift#L549): `updateBottomInsetForBottomBarHeight()` sets `self.view.keyboardLayoutGuide.keyboardDismissPadding = bottomBarHeight` without validating `bottomBarHeight`. If `bottomBarHeight` becomes negative (e.g., layout glitch or incorrect measurement), this can produce a negative dismiss padding, creating an invalid drag-to-dismiss region and unpredictable gesture behavior. Clamp to non-negative before assignment, e.g., `keyboardDismissPadding = max(0, bottomBarHeight)`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Messages/Messages View Controller/View Controller/Views/MessagesBottomBar.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 73](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/Messages View Controller/View Controller/Views/MessagesBottomBar.swift#L73): `onChange(of: focusCoordinator.currentFocus)` will not reliably fire because `focusCoordinator` is a plain `let` reference (not a SwiftUI state/ObservedObject) and `onChange` only detects value changes between body evaluations. If `currentFocus` changes internally without causing this view to re-render, the closure won't be invoked and `isExpanded` won't update, leaving the UI stuck in the wrong state. Use a SwiftUI-observed source (e.g., make `FocusCoordinator` an `ObservableObject` and store it as `@ObservedObject/@StateObject`, or forward an `@Binding`/state value) or switch to a publisher-based observation. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversation Detail/Messages/Messages View Controller/View Controller/Views/MessagesInputView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 59](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/Messages View Controller/View Controller/Views/MessagesInputView.swift#L59): The TextField placeholder uses `profile.displayName` directly: `"Chat as \(profile.displayName)"`, ignoring both `displayName` (the bound, potentially user-edited value) and `emptyDisplayNamePlaceholder`. If `profile.displayName` is empty, the placeholder becomes incomplete, and the provided `emptyDisplayNamePlaceholder` input is never used, causing externally visible behavior that likely contradicts the component’s API. Consider using `displayName.isEmpty ? emptyDisplayNamePlaceholder : displayName`. <b>[ Out of scope ]</b>
- [line 71](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversation Detail/Messages/Messages View Controller/View Controller/Views/MessagesInputView.swift#L71): `onSubmit` sends the message unconditionally, ignoring `sendButtonEnabled`. This allows submitting via Return/Submit even when the Send button is disabled (e.g., empty or invalid input), creating a contract mismatch and potential unintended sends. Add a guard: `if sendButtonEnabled { onSendMessage() }` or replicate the same validation used to compute `sendButtonEnabled`. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Conversations List/ConversationsListEmptyCTA.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 46](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversations List/ConversationsListEmptyCTA.swift#L46): `URL(string: "https://xmtp.org")!` is force-unwrapped in the `openURL` call. If the string ever becomes malformed, `URL(string:)` will return `nil` and the force unwrap will crash at runtime when the button is tapped. Replace with an `if let url = URL(string: ...) { openURL(url, prefersInApp: true) }` guard or pre-validated constant. <b>[ Out of scope ]</b>
- [line 67](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversations List/ConversationsListEmptyCTA.swift#L67): `URL(string: "https://convos.org/terms-and-privacy")!` is force-unwrapped in the `openURL` call. If the string ever becomes malformed, `URL(string:)` will return `nil` and the force unwrap will crash at runtime when the button is tapped. Use an `if let` guard or validated constant to avoid crashes. <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/Conversations List/ConversationsView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 118](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversations List/ConversationsView.swift#L118): Duplicate transition source IDs: `matchedTransitionSource(id: "composer-transition-source", in: namespace)` is applied to two different toolbar items (the “Scan” button at lines 118–121 and the “Compose” button at lines 128–131). SwiftUI’s matched transitions expect a unique source per ID; duplicating the ID can yield ambiguous or incorrect animation source resolution (e.g., the zoom transition for `NewConversationView` may originate from the wrong control or behave inconsistently). Use distinct IDs per control or only tag the actual source control for the destination. <b>[ Out of scope ]</b>
- [line 128](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversations List/ConversationsView.swift#L128): Misleading transition mapping: The “Scan” button declares `matchedTransitionSource(id: "composer-transition-source", in: namespace)` even though its action (`viewModel.onJoinConvo()`) does not present the composer `fullScreenCover`. This can cause the `NewConversationView` zoom transition to originate from the Scan button if SwiftUI resolves that source, resulting in an incorrect visual association. Remove the tag from Scan or give each control its correct, unique transition source ID. <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/Conversations List/ConversationsViewModel.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 57](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Conversations List/ConversationsViewModel.swift#L57): `selectedConversationId` setter silently clears `_selectedConversationId` and `selectedConversation` when the provided `newValue` does not match any entry in `conversations` (lines `57`–`60`). This discards the caller-provided ID and loses the ability to resolve selection later when `conversations` are populated/updated. If callers set an ID before data arrives or while the list is transiently out of sync, selection is unexpectedly reset to `nil`. Consider retaining the unmatched ID in `_selectedConversationId` and deferring resolution of `selectedConversation` until the conversation list contains it, or explicitly rejecting the input with an error/result. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Shared Views/ConversationInfoPresenter.swift — 0 comments posted, 3 evaluated, 2 filtered</summary>

- [line 49](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/ConversationInfoPresenter.swift#L49): `onChange(of: focusCoordinator.currentFocus)` is unlikely to ever fire because `focusCoordinator` is a plain `let` reference and is not an observed SwiftUI state (`@StateObject`, `@ObservedObject`, `@Bindable`, etc.). SwiftUI only re-evaluates `body` (and therefore detects changes for `onChange`) when observed state changes. Mutations inside `focusCoordinator` will not trigger a view update, so changes to `currentFocus` will not be detected and `focusState` will not be synchronized. This silently breaks the intended focus synchronization path. <b>[ Low confidence ]</b>
- [line 55](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/ConversationInfoPresenter.swift#L55): A bidirectional synchronization loop risk exists between `onChange(of: focusCoordinator.currentFocus)` which sets `focusState = newFocus` and `onChange(of: focusState)` which calls `focusCoordinator.syncFocusState(newFocus)`. If `syncFocusState(_:)` updates `currentFocus` to a different value or toggles flags that cause a different `currentFocus`, this can create a repeated feedback loop (re-entrancy) and oscillate focus changes. Without atomic swap-and-clear or idempotency guarantees, this can lead to unstable focus behavior. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Shared Views/ConversationToolbarButton.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>

- [line 56](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/ConversationToolbarButton.swift#L56): Using `.fixedSize()` on the `Text(title)` conflicts with `.lineLimit(1)` and `.truncationMode(.tail)`. `fixedSize()` prevents truncation/compression, so long titles may overflow/clamp the 180pt frame without showing the expected ellipsis, causing clipping/overlap in the `HStack`. Consider removing `.fixedSize()` or relying on the frame and truncation to ensure proper ellipsis. <b>[ Out of scope ]</b>
</details>

<details>
<summary>Convos/Shared Views/FocusCoordinator.swift — 0 comments posted, 10 evaluated, 6 filtered</summary>

- [line 308](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/FocusCoordinator.swift#L308): `updateKeyboardState(frame:isShowEvent:screen:)` contradicts its own comment: it says "Use the screen from the notification, or fall back to main screen", but it requires a non-nil `screen` parameter and never falls back inside the function. Combined with delegate methods that return early on `nil` screen, keyboard detection is effectively disabled on typical iOS. Implement a fallback (`let screen = screen ?? UIScreen.main`). <b>[ Low confidence ]</b>
- [line 354](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/FocusCoordinator.swift#L354): `updateKeyboardState(frame:isShowEvent:screen:)` only updates `currentFocus` when `currentFocus == nil` (line ~355), but the preceding comment explicitly states it should update when `currentFocus` is `nil` or `.message`. This mismatch can leave focus at `.message` when it should adjust to the new default after keyboard type changes. Update the condition to `if currentFocus == nil || currentFocus == .message { ... }`. <b>[ Low confidence ]</b>
- [line 374](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/FocusCoordinator.swift#L374): The delegate method `keyboardDidShow(info:)` returns early when `info.screen` is `nil`, so the keyboard "show" event never updates detection. Given `notification.object` is commonly `nil`, this blocks updates in typical scenarios. Provide a fallback (e.g., `UIScreen.main`). <b>[ Low confidence ]</b>
- [line 381](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/FocusCoordinator.swift#L381): The delegate method `keyboardWillHide(info:)` returns early when `info.screen` is `nil`, preventing state updates on hide. This leaves `keyboardType` unchanged and can break subsequent `defaultFocus` logic. Add a fallback to `UIScreen.main`. <b>[ Low confidence ]</b>
- [line 390](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/FocusCoordinator.swift#L390): The delegate method `keyboardDidHide(info:)` returns early when `info.screen` is `nil`, blocking keyboard state updates on hide and leaving `keyboardType` stale. Use `UIScreen.main` when `screen` is absent. <b>[ Low confidence ]</b>
- [line 410](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/FocusCoordinator.swift#L410): The delegate method `keyboardDidChangeFrame(info:)` returns early when `info.screen` is `nil`, preventing detection of show/hide via frame changes. Add a fallback to `UIScreen.main` to compute visibility and update state. <b>[ Low confidence ]</b>
</details>

<details>
<summary>Convos/Shared Views/QuickEditView.swift — 0 comments posted, 2 evaluated, 2 filtered</summary>

- [line 10](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/QuickEditView.swift#L10): `textFieldShouldReturn(_:)` returns `false`, which prevents the default return handling (such as resigning first responder). Combined with invoking `action?()` without any explicit focus change, pressing Return may leave the keyboard up and the field focused after submission, degrading UX and potentially trapping focus. If the intent is to allow default handling, return `true` after calling the action, or explicitly manage focus state (e.g., clear `focusState`) to produce a defined terminal state. <b>[ Low confidence ]</b>
- [line 35](https://github.com/ephemeraHQ/convos-ios/blob/e30fd3c4ddb992018a751c04078b08c1c89751e9/Convos/Shared Views/QuickEditView.swift#L35): `onSubmit` is only wired via UIKit delegate inside `.introspect(.textField, on: .iOS(.v26))` and there is no `.onSubmit {}` SwiftUI handler. On iOS versions not matching `.v26`, the introspection closure will not run, so pressing the Return/Done key will not invoke `onSubmit`, producing inconsistent submission behavior across OS versions and potentially no-op on submit. Add `.onSubmit { onSubmit() }` to the `TextField` to ensure consistent behavior, and keep the delegate as an optional enhancement. <b>[ Low confidence ]</b>
</details>

</details>


yewreeka:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/239?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#240** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/240?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#239** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/239?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/239?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#238** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/238?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#231** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/231?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=239).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #670, opened 2026-04-08
Fix incoming video aspect ratio when no thumbnail or stored dimensions

yewreeka:
The previous fix for incoming video aspect ratio only ran when the
attachment had a thumbnail — it decoded the thumbnail and forwarded its
dimensions via onDimensionsLoaded. That covered outgoing videos (which
the iOS sender always compresses and thumbnails) and remote attachments
with a sender-provided thumbnail, but it missed two real cases:

1. Inline video attachments from clients that don't populate
   StoredRemoteAttachment.mediaWidth / mediaHeight / thumbnailDataBase64
   (e.g. the convos CLI's send-attachment command for files under 1 MB.
2. Remote video attachments that arrive without a thumbnail.

For both cases HydratedAttachment.width / height stayed nil forever,
aspectRatio returned nil, and the bubble fell back to a 4:3 placeholder.
Sending a 360x640 portrait video from the CLI reproduced this as a
squat landscape rectangle instead of the expected tall portrait frame.

Fix: once the decrypted video file is available on disk, read the
natural video dimensions from the file itself and feed them through
the same onDimensionsLoaded path the thumbnail branch already uses.
VideoCompressionService gains a loadVideoDimensions(from:) helper that
reuses the existing AVURLAsset + preferredTransform logic (so rotated
portrait captures return portrait dimensions). AttachmentPlaceholder
calls it only when attachment.width is still nil after the thumbnail
branch, so there's no extra AVAsset work when dimensions are already
available.

Verified on simulator by sending /tmp/test-cat.mp4 (360x640) via the
convos CLI: bubble now renders as 9:16 portrait and the dimensions
persist to AttachmentLocalState for subsequent renders.
EOF
)

> [!NOTE]
> ### Fix incoming video aspect ratio when no thumbnail or stored dimensions
> - Adds `loadVideoDimensions(from:)` to `VideoCompressionService` that reads a video track's `naturalSize` and `preferredTransform` via AVFoundation, returning the correct dimensions accounting for rotation.
> - In `MessagesGroupItemView`, when a video attachment has no pre-populated width/height, the view now calls this method after resolving the local/remote URL and reports results via `onDimensionsLoaded` on the main actor.
> - Adds a test asserting `loadVideoDimensions` throws `invalidAsset` for non-video data.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 05ffd40.</sup>
> 
>

### DIFF
--- a/Convos/Conversation Detail/Messages/MessagesListView/MessagesGroupItemView.swift
+++ b/Convos/Conversation Detail/Messages/MessagesListView/MessagesGroupItemView.swift
@@ -607,6 +607,9 @@ private struct AttachmentPlaceholder: View {
 
             do {
                 let videoURL = try await resolveVideoURL(for: attachment.key)
+                if attachment.width == nil {
+                    await loadVideoDimensionsIfPossible(from: videoURL)
+                }
                 let player = AVPlayer(url: videoURL)
                 await player.seek(to: .zero)
                 inlinePlayer = player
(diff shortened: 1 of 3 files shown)


yewreeka:
* **#670** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/670?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/670?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=670).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #720, opened 2026-04-21
Remove new-engineer onboarding friction

yewreeka:
## Summary

Walking through `README.md` + `Scripts/setup.sh` as a new engineer on a fresh machine turned up several papercuts. This PR fixes them.

### README.md
- Drop `Ruby 3.3.3` from prerequisites — no `Gemfile` exists in the repo, and `setup.sh` never ran `bundle install`.
- Add `Docker` to prerequisites — required by `./dev/test` and the XMTP local node, but it was only mentioned in `CLAUDE.md`.
- Expand **Quick Start** to include `cp .env.example .env` and the optional Firebase App Check debug token step.
- Fix the broken test example: the scheme `Convos` doesn't exist; use `Convos (Local)` with a `-destination`.
- Document the real test path for most devs: `./dev/test` (or `./dev/up` + `swift test --package-path ConvosCore` + `./dev/down`).
- Add **Useful Make Targets** (`make setup`/`status`/`secrets-local`/`protobuf`/`clean`) so folks can find them.
- Link to `ENVIRONMENTS.md` from the setup section.

### Scripts/setup.sh
- **Worktree-safe git hooks**: replace the per-hook symlinks into `.git/hooks/` with `git config core.hooksPath Scripts/hooks`. The old approach `cd`-ed into `${repo}/.git`, which is a **file** (not a directory) inside any `git worktree` — so it silently failed or pointed at the wrong place in `convos-task` worktrees.
- **No-sudo SwiftLint install**: install the pinned `portable_swiftlint` to `$(brew --prefix)/bin` instead of `sudo mv …  /usr/local/bin`. Avoids the mid-run password prompt, works on both Intel and Apple Silicon without PATH gymnastics.
- **Xcode version check**: `die` early if `xcodebuild -version` reports Xcode < 16, matching the README's stated minimum.
- **Better missing-`.env` hint**: the previous message said "create one" but didn't mention that `.env.example` is already a 4.8 KB template in the repo. Now it tells you to `cp .env.example .env`.

## Test plan
- [x] Run `./Scripts/setup.sh` on a fresh clone — completes cleanly, no sudo prompt
- [x] Verify `git config core.hooksPath` is set to `Scripts/hooks`
- [x] Make a commit → `pre-commit` hook fires (`🔍 Running pre-commit checks...`)
- [x] Push the branch → `pre-push` SwiftLint runs across all 510 Swift files with 0 violations
- [ ] Reviewer: try `./Scripts/setup.sh` from inside a `convos-task` worktree and confirm hooks fire there too
- [ ] Reviewer: sanity-check the updated `xcodebuild test` command runs locally

> [!NOTE]
> ### Reduce new-engineer onboarding friction by updating setup script and README
> - [setup.sh](https://github.com/xmtplabs/convos-ios/pull/720/files#diff-490f2d9bc338440322d4b7d5bcb8c209ee7633cdb9f4245ceb2cb78bc6486314) now configures Git hooks via `core.hooksPath` pointing to `Scripts/hooks` instead of per-hook symlinks in `.git/hooks`, and adds an Xcode 16+ version check that aborts setup if the requirement isn't met.
> - SwiftLint is installed into the Homebrew prefix bin directory without `sudo`, replacing the previous `/usr/local/bin` destination.
> - [README.md](https://github.com/xmtplabs/convos-ios/pull/720/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) expands the Quick Start with Docker prerequisites, a `make setup` option, `.env` setup from template, and a list of useful `make` targets; testing instructions now prefer `./dev/test` with a Docker-based XMTP node.
> - Behavioral Change: existing clones that previously relied on `.git/hooks` symlinks will need to re-run `setup.sh` to pick up the new `core.hooksPath` configuration.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9dc3dc5.</sup>
> 
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,46 +22,41 @@ Learn more at [convos.org](https://convos.org).
 ### Prerequisites
 
 - macOS with Xcode 16+
-- Ruby 3.3.3
 - Homebrew
+- Docker (required for running the test suite via `./dev/up`)
 
 ### Quick Start
 
 ```bash
-# Run the setup script
-./Scripts/setup.sh
+# 1. Install dependencies and configure your environment
+./Scripts/setup.sh        # or: make setup
+
+# 2. Create your local .env from the template
+cp .env.example .env
+
+# 3. (Optional) Add a Firebase App Check debug token to .env
+#    See .env comments, or generate one at:
+#    https://console.firebase.google.com/project/convos-otr/appcheck
 ```
 
 The setup script will:
-- Install Git hooks for code quality checks
+- Configure the repo's Git hooks (`core.hooksPath` → `Scripts/hooks/`), which works for both regular clones and `git worktree`s
 - Configure Xcode defaults for consistent development
 - Install required dependencies:
-  - SwiftLint (code linting)
+  - SwiftLint (code linting, pinned version for compatibility)
   - SwiftFormat (code formatting)
   - swift-protobuf (Protocol Buffer support)
   - GitHub CLI (for release automation)
-  - Bundler and Ruby gems
 
 ### Manual Setup
 
-If you prefer to set up components individually:
+If you prefer to install dependencies individually, run:
 
 ```bash
-# Install SwiftLint
-brew install swiftlint
-
-# Install SwiftFormat
-brew install swiftformat
-
-# Install swift-protobuf
-brew install swift-protobuf
-
-# Install GitHub CLI (optional, for releases)
-brew install gh
-
-# Install Ruby dependencies
-bundle install
+brew install swiftformat swift-protobuf gh
 ```
+
+Then run `./Scripts/setup.sh` once to install the pinned SwiftLint version, configure git hooks, and apply Xcode defaults.
 
 ### Configuration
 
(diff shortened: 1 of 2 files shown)


yewreeka:
Validated the worktree scenario end-to-end in a fresh clone (`convos-ios-task-D`):

```bash
git config core.hooksPath Scripts/hooks            # the fix
git worktree add .claude/worktrees/hookpath-test   # the scenario that used to break
ls -la .claude/worktrees/hookpath-test/.git
# → -rw-r--r-- … 82 bytes  (a *file*, not a directory)

cd .claude/worktrees/hookpath-test
echo x > x.txt && git add x.txt && git commit -m 'test'
# → 🔍 Running pre-commit checks...
# → ✓ No Swift files staged, skipping checks
# → [hookpath-test c8f29457] test
```

Both `post-checkout` (on `git worktree add`) and `pre-commit` hooks fired correctly from the tracked `Scripts/hooks/` directory — which is exactly the thing the old `.git/hooks/*` symlinks couldn't do inside a worktree.

PR #725, opened 2026-04-22
feat: inactive-conversation mode for post-restore UX

yewreeka:
Adds the "Awaiting reconnection" state that floats ahead of the backup
stack as PR 2 of docs/plans/icloud-backup-single-inbox.md. The feature
is useful on its own for any network-recovery scenario where the MLS
group temporarily can't accept writes, and de-risks the restore stack
by landing the data-layer + UI pieces before RestoreManager needs them.

Data layer
- ConversationLocalState gains an isActive column (v2-inactive-conversations
  migration, default true so existing rows are unaffected).
- ConversationLocalStateWriter gains setActive(_:for:) and
  markAllConversationsInactive() (bulk UPDATE in a single write txn for
  RestoreManager).
- Conversation model surfaces isActive: Bool, threaded through hydration.
- DBMessage.Update gains isReconnection: Bool with a backward-compatible
  decoder (defaults to false when the key is missing so legacy rows
  continue to deserialize).

Reactivation in StreamProcessor
- markReconnectionIfNeeded flips isActive = true when a message arrives on
  an inactive conversation AND the MLS group sync succeeded. A failed
  sync ("group is inactive") leaves the flag alone because the
  installation can't participate yet.
- markRecentUpdatesAsReconnection backfills isReconnection on the 5 most
  recent update rows so the UI renders the reconnection point as a
  visible event rather than a cliff.
- conversationWriter.store is wrapped in a fallback helper that falls
  back to the existing DBConversation row on sync failure but flags
  syncFailed so the reactivation check is skipped.

UI
- New InactiveConversationBanner: "Restored from backup" pill above the
  composer, tappable to learn.convos.org.
- ConversationView intercepts send/reaction/toggleReaction/reply on
  inactive conversations and shows an "Awaiting reconnection" alert.
- ConversationsListItem subtitle shows "Awaiting" next to the date,
  matching the existing isPendingInvite / "Verifying" pattern.

Tests
- ConversationLocalStateWriterTests (5 cases) exercise setActive and
  markAllConversationsInactive happy paths + scope isolation.
- DBMessageUpdateDecodingTests (4 cases) lock in backward-compatible
  decoding (missing key → false), round-trip, and memberwise default.

Validation
- swift test --package-path ConvosCore: 583/583 pass.
- swiftlint: 0 violations on StreamProcessor after extracting the
  sync-fallback helper.

Plan: docs/plans/icloud-backup-single-inbox.md (PR #724).

> [!NOTE]
[...]

PR #736, opened 2026-04-23
feat(backup): bundle format, crypto, metadata

yewreeka:
feat(backup): bundle format, crypto, metadata

First checkpoint of PR 3 (iCloud backup port). Self-contained: no
callers yet, just the leaf pieces restore and backup will build on.

- BackupBundle: tar pack/unpack with "CVBD" magic + 1-byte version
  header at the head of the tar. Future bundle-format bumps can
  discriminate cleanly without having to decode past-incompatible
  data. Path-traversal hardening (standardized-path check + symlink-
  resolved re-check after createDirectory) ported from the old
  louis/icloud-backup branch unchanged.
- BackupBundleCrypto: AES-256-GCM seal/open keyed directly on the
  identity's databaseKey. No HKDF. Also generates per-bundle 32-byte
  archiveKeys for the inner XMTP archive — fresh CSPRNG, never
  reused across bundles.
- BackupBundleMetadata: full vs sidecar projections. Full form lives
  inside the encrypted tar and carries the inner archiveKey; sidecar
  form sits unencrypted next to backup-latest.encrypted for discovery
  and strictly omits every secret. One Codable type with a secret-free
  nested Sidecar struct reachable via a computed property.

Tests: 9 cases
- Round-trip: pack → unpack restores all entries byte-for-byte.
- Header: rejects missing CVBD magic; rejects unsupported format
  versions (forged 0xFF in the version byte).
- Path traversal: hand-crafted tar with "../escape.bin" entry is
  refused and no file lands outside the staging dir.
- Crypto: rejects non-32-byte keys on both encrypt and decrypt;
  consecutive generateArchiveKey() calls don't collide.
- Metadata: full round-trip preserves archiveKey; sidecar omits
  archiveKey (verified by searching the serialized JSON); readSidecar
  works on both sidecar and full files; readFull fails when
  archiveKey is missing (legacy decoder path).

Plan: docs/plans/icloud-backup-single-inbox.md §"Bundle format + crypto"
and §"Two keys, two roles — don't collapse them".

feat(backup): DatabaseManager.replaceDatabase + NSE restore gate

Second checkpoint of PR 3. Adds the destructive-swap primitive
RestoreManager will drive on top, plus the app-group flag that
keeps the NotificationService Extension out of the way while the
DB is being rewritten page-by-page.

DatabaseManager
- replaceDatabase(with backupPath: URL) on DatabaseManagerProtocol.
  Pool-to-pool copy via GRDB's backup(to:) so the DatabasePool
  instance itself isn't swapped — any long-lived reader/writer
  elsewhere stays valid across the restore.
- Captures a pre-restore snapshot first; if the copy throws, we
  replay the snapshot back into the live pool and re-migrate.
  Double-fault (copy throws AND rollback throws) surfaces as
  DatabaseManagerError.rollbackFailed, which the UI must treat
  as fatal.
- WAL checkpoint (TRUNCATE) before the swap so the on-disk file
  is a consistent self-contained snapshot; any coordinated
  reader after the barrier sees the new contents, not a mixture
  of new pages + stale WAL tail.
- Entire operation runs under NSFileCoordinator.coordinate(
  writingItemAt:,options: .forReplacing) against the DB URL —
  the NSE, if it happened to be the only reader not honoring the
  flag, would block at the coordinator barrier instead of
  tearing a read.

RestoreInProgressFlag
- New helper at ConvosCore/Backup/RestoreInProgressFlag.swift.
- Lives in app-group UserDefaults under suite
  "convos.restore-in-progress" + key
  "convos.restore-in-progress.flag".
- Set/cleared by the yet-to-land SessionManager pauseForRestore /
  resumeAfterRestore (Checkpoint 3). Checked by the NSE at entry.
- Plan §"NSE coordination during DB swap" — app-group flag is
  belt, NSFileCoordinator is suspenders; either alone is
  insufficient.

NotificationService
- didReceive early-exits with an empty content delivery when the
  flag is set. Push loss within a narrow user-initiated restore
  window is an acceptable trade for not risking a torn read on
  the shared GRDB.

MockDatabaseManager
- Mirrors the real replaceDatabase contract (rollback snapshot +
  migrate) for test fixtures. Skips NSFileCoordinator (single-
  process) and the WAL checkpoint (in-memory queues don't WAL).

Tests (3 cases, DatabaseManagerReplaceTests)
- Successful replace: live pool reflects backup's rows after
  replaceDatabase.
- Missing backup file: throws DatabaseManagerError.backupFileMissing
  before touching the live pool.
- Corrupt backup: rolls back; live pool returns to the pre-restore
  state (verified by row contents, not just the throw).

Plan: docs/plans/icloud-backup-single-inbox.md §"DatabaseManager.
replaceDatabase + NSE coordination" and §"Throwaway XMTP client
for archive import" (the in-process flag arrives in Checkpoint 3
as SessionManager's isRestoringInProcess).

fix(backup): tighten CP1/CP2 loose ends before CP3

Small correctness + clarity fixes surfaced by architect review of
the bundle-format and replaceDatabase checkpoints. No behavior
change on the happy path; closes a minor edge case, makes the
restore-flag failure mode strict, and consolidates a file-path
constant so the CP3 backup writer and DatabaseManager can't drift.

- BackupBundle.untarData: guard pathLength > 0. An empty path would
  decode to the empty string and resolve via appendingPathComponent("")
  to the staging directory URL itself. The containment check already
  catches that (prefix match fails), but an explicit early bail gives
  a clearer error and keeps the intent visible.
- BackupBundleMetadata.readSidecar: add a code comment pinning
  "JSONDecoder's default ignore-unknown-keys behavior is load-bearing"
  so a future switch to a strict decoder doesn't silently break
  reading full-form files.
- RestoreInProgressFlag.set is now strict (throws on missing
  app-group container) while read stays lenient (degrades to false).
  Asymmetric on purpose: the NSE reading a stale false value
  degrades to current behavior, whereas a silent set failure would
  let pauseForRestore proceed into destructive ops with an
  un-signaled NSE. SessionManager.pauseForRestore (CP3) must abort
  on throw.
- DatabaseManager.databaseFilename exposed as a public static
  constant; BackupBundle.Component.database now references it.
  Single source of truth for "convos-single-inbox.sqlite". CP3's
  BackupManager will pick up the same constant.

One architect finding dismissed: SharedDatabaseMigrator's
eraseDatabaseOnSchemaChange is already #if DEBUG-gated (lines
19-21), contrary to the agent's reading. Rollback path is safe
in release builds as designed.

Tests: 12/12 pass across BackupBundleTests (now 9 cases) and
DatabaseManagerReplaceTests (3 cases). No new cases needed — the
edge case guard in untarData is a belt-and-suspenders on top of
the existing path-traversal test.

feat(backup): SessionManager pause/resume for restore lifecycle

CP3a of PR 3 — the coordination seam `RestoreManager` will drive.
No callers yet inside the repo; the next commit (CP3b) adds the
archive wrappers + revoker, and CP3c/d wire the actual managers.

SessionManager
- New `isRestoringInProcess: Bool` flag, lock-guarded via the same
  `cachedMessagingService` OSAllocatedUnfairLock as the cache slot
  itself. Strict pairing matters: a concurrent `loadOrCreateService()`
  must see either both (restoring + cleared slot) or neither
  (not-restoring + live cache).
- `loadOrCreateService()` short-circuits when the flag is set:
  returns a frozen `MessagingService(identityReadFailure:
  RestoreInProgressError())` placeholder. Cached under the lock so
  repeated `messagingService()` calls during pause don't thrash.
  The error-state observer sees `.error(RestoreInProgressError)`
  and can render "Restoring…" without attempting recovery.
- Package-internal `pauseForRestore()` / `resumeAfterRestore()`:
  - Pause: sets `RestoreInProgressFlag` (app-group) strict-throws,
    flips the in-process flag inside the lock, calls
    `unusedConversationCache.cancel()`, stops (not deletes) the
    existing cached service, clears the slot. On throw from the
    flag-set step, RestoreManager must abort before any destructive
    op — the NSE coordination contract depends on both flags.
  - Resume: clears the in-process flag + slot atomically, clears
    the app-group flag (log-on-failure since we're already
    committed past the destructive point), triggers
    `loadOrCreateService()` to rebuild against the restored DB.

RestoreInProgressError
- Surfaces via `SessionStateMachine.State.error(RestoreInProgressError)`.
- Not a `TerminalSessionError`: this state clears when resume
  completes.

Tests (4 cases, SessionManagerRestoreTests, .serialized):
- pauseForRestore sets the app-group flag; resume clears it.
- messagingService() during pause returns a RestoreInProgressError
  placeholder.
- Placeholder is cached — repeated calls return same instance.
- Resume evicts placeholder; next call builds a fresh service.

Suite is `.serialized` because all tests share the app-group
UserDefaults suite that RestoreInProgressFlag lives in — parallel
execution would race on flag reads/writes.

Plan: docs/plans/icloud-backup-single-inbox.md §"Throwaway XMTP
client for archive import" + §"Restore integration with
SessionManager directly".

feat(backup): XMTP archive protocol surface + installation revoker

CP3b of PR 3. Extends the XMTPClientProvider seam with the archive
and inbox-state methods that BackupManager/RestoreManager will
drive in CP3c/d, plus the installation revoker that runs at the
tail of a restore. Also lands the two architect-review fixes to
CP3a's SessionManager pause/resume.

XMTPClientProvider additions (protocol + Client extension + mocks)
- createArchive(path:encryptionKey:) — forwards to XMTPiOS.Client
  with default ArchiveOptions. Explicit pass-through required
  because the SDK's method takes a third `opts` arg with default;
  Swift's protocol witness doesn't auto-bridge that shape.
- importArchive(path:encryptionKey:) — signature matches the SDK
  exactly, default witness resolves.
- inboxState(refreshFromNetwork:) — needed by the revoker to
  enumerate live installations. refreshFromNetwork: true forces a
  fresh read; the cached view is often stale post-importArchive.

Mock + TestableMockClient conformances updated. MockXMTPClientProvider
grows call-recording for createArchive / importArchive /
revokeInstallations so tests can assert the protocol calls happened
without spinning up a real XMTP client. stubInboxState + inboxStateError
let tests drive the inboxState branch.

XMTPInstallationRevoker (new)
- Static util taking any XMTPClientProvider. Fetches inbox state
  (refreshFromNetwork: true), filters out keepInstallationId,
  revokes the rest. Returns the revoked count. Throws
  .noActiveInstallations if the network refresh yields zero
  installations (expected-impossible post-import, but guard
  catches SDK weirdness before we hand an empty list to
  revokeInstallations).
- Protocol-seam design per architect call: keeps the revoker
  mockable without a real Client. No mock test in this commit —
  XMTPiOS.InboxState has no public initializer outside the SDK,
  so the revoker is exercised via RestoreManager's integration
  path in CP3d.

SessionManager architect fixes to CP3a
- pauseForRestore: post-flag work wrapped in do/catch with
  Task.checkCancellation() at the top so the catch is reachable.
  Any throw (cancellation today, future cancel/stop throws) now
  runs resumeAfterRestore() and rethrows — no more half-paused
  sessions on partial failure.
- resumeAfterRestore: dropped the eager loadOrCreateService()
  call. Lazy rebuild matches tearDownInbox's pattern (nil the
  slot, let the next caller rebuild). Previously, a
  keychain-unreadable-post-restore error would silently cache a
  FailedIdentityLoadOperation-backed service and return void,
  hiding the causal link for RestoreManager.

Deferred: per-test appGroupIdentifier suffix to replace the
.serialized scope in SessionManagerRestoreTests. Fine for now;
revisit at CP3c when more suites touch RestoreInProgressFlag.

Dismissed from architect review: "eraseDatabaseOnSchemaChange
unconditional" — verified still #if DEBUG-gated at
SharedDatabaseMigrator.swift:19-21.

Tests: 16/16 pass across touched suites
(SessionManagerRestoreTests 4, BackupBundleTests 9,
DatabaseManagerReplaceTests 3). SwiftLint clean on all files.

Plan: docs/plans/icloud-backup-single-inbox.md §"Throwaway XMTP
client for archive import" — the three invariants' enforcement
lands in RestoreManager (CP3d); the protocol-level primitives
ship here.

feat(backup): BackupManager — sealed-bundle create flow

CP3c of PR 3. Lands the backup-side manager that BackupScheduler
(CP5) and the "Back up now" button will drive. Read-path matches
the plan's §"Backup flow".

BackupManager (new)
- Non-actor final class. createBackup() is inherently serialized
  by its one caller (scheduler fires sequentially, UI is user-
  driven); concurrency protection lives at the call site.
- Flow: skip-if-restore-in-progress → loadSync identity (skip if
  nil) → GRDB snapshot via dbReader.backup(to:) → fresh 32-byte
  archiveKey → client.createArchive via the CP3b protocol seam →
  buildMetadata (conversationCount, schemaGeneration, appVersion,
  archiveKey inner) → writeFull into the staging tar → pack + seal
  with identity.keys.databaseKey → atomic write to iCloud Documents
  /backups/<deviceId>/backup-latest.encrypted, fallback to local.
- Sidecar written *after* the sealed bundle so findAvailableBackup
  readers can't see a sidecar pointing at a not-yet-landed bundle.
  The sidecar omits archiveKey (secret stays inside the seal).
- BackupError surface: .noIdentity, .restoreInProgress,
  .clientUnavailable, .archiveFailed, .writeFailed — each rooted at
  the specific failure domain so callers can branch on the semantic,
  not parse localized strings.
- Staging dir cleanup via `defer { BackupBundle.cleanup(...) }` so
  cancelled/failed backups never leak bytes.
- `now:` closure in the init is a test seam — BackupManagerTests
  pin the createdAt deterministically without freezing wall-clock.

Prerequisite additions
- AppEnvironment.iCloudContainerIdentifier — derived from the
  app-group identifier (same pattern the old louis/icloud-backup
  branch used); Rev 4 plan §"Per-device backup path" assumes this
  accessor exists.
- DeviceInfoProviding.deviceName — new protocol requirement + iOS
  impl (captures UIDevice.current.name at init) + MockDeviceInfoProvider
  default. Backed by the user-facing requirement in the plan's
  fresh-install restore card: "A backup from [deviceName] is
  available". deviceIdentifier is stable-but-opaque; deviceName is
  display-facing.

Tests (4 cases, BackupManagerTests, .serialized)
- Happy path: createBackup returns a URL, file exists, sealed bundle
  unpacks with the identity's databaseKey, full metadata inside the
  tar carries archiveKey, MockXMTPClientProvider.createArchive was
  called once with that exact archiveKey, sidecar next to the bundle
  omits archiveKey (verified via JSON substring search).
- skips with .restoreInProgress — MockXMTPClientProvider never gets
  called (assertion on createArchiveCalls.isEmpty).
- skips with .noIdentity — same isEmpty assertion.
- archive failure surfaces as BackupError.

Design notes
- clientProvider is a `@Sendable () async throws -> any XMTPClientProvider`
  closure in the init. Callers that have a live SessionManager bind
  `{ try await session.messagingService().waitForInboxReadyResult().client }`.
  Tests bind a trivial `{ mockClient }`. No production adapter
  protocol, no production mock — one seam.
- conversationCount query filters `isUnused = 0` so unused-cache
  prewarm rows don't inflate the sidecar count shown in the restore
  prompt.
- The sidecar-after-bundle write ordering is half of the
  "pair-atomic write" concern the architect raised in the CP1/CP2
  review; the other half (findAvailableBackup cross-check) lands
  with RestoreManager in CP3d.

Plan: docs/plans/icloud-backup-single-inbox.md §"Backup flow (new)".

feat(backup): RestoreManager — decrypt, swap DB, import archive, revoke

CP3d of PR 3 — the restore side. Companion to BackupManager (CP3c).
The full flow from the plan's §"Restore flow" in a single actor,
with the throwaway-XMTP-client contract (three invariants from
plan §"Throwaway XMTP client for archive import") enforced
in-process.

RestoreManager (new)
- Actor. restoreFromBackup(bundleURL:) is the one entry point.
  Public findAvailableBackup(environment:) static for discovery
  that doesn't need the bundle key (sidecar read only).
- Wired dependencies: DatabaseManagerProtocol (replaceDatabase),
  KeychainIdentityStoreProtocol (loadSync polling),
  SessionManager (pauseForRestore / resumeAfterRestore), and a
  ThrowawayClientBuilder closure that hands out an
  XMTPClientProvider on demand. The production default builder
  calls XMTPiOS.Client.build against the restored DB directory;
  tests inject a closure that returns MockXMTPClientProvider.

The flow executes the Rev 4 plan literally:
  1. Read + decrypt bundle (seal key = identity.databaseKey).
  2. awaitIdentity — bounded loadSync poll (default 30s, 500ms
     poll) so iCloud Keychain lag doesn't cause
     `loadOrCreateService` to `.register` a forked identity.
  3. Validate metadata.schemaGeneration against currentGeneration;
     mismatch throws .schemaGenerationMismatch (distinct from
     decryption / validation errors so the UI can render the
     user-facing "made on an older version" copy + emits a
     QAEvent telemetry hit for future diagnostics).
  4. Validate GRDB snapshot + xmtp-archive.bin + archiveKey
     present before any destructive op.
  5. SessionManager.pauseForRestore() — sets both app-group
     and in-process flags, stops the cached service, clears
     the slot. Throws wrap in .sessionPauseFailed.
  6. Stage xmtp-*.db3 files aside in a temp dir (rollback
     anchor).
  7. DatabaseManager.replaceDatabase (WAL-checkpoint +
     NSFileCoordinator write barrier, from CP2).
  8. Build throwaway XMTP client via clientBuilder →
     importArchive → dropLocalDatabaseConnection in defer.
     Any throw here is CAUGHT and recorded as
     .archiveImportFailed — GRDB restore is already
     committed, so degrading to conversation-list-only is
     strictly better than aborting.
  9. markAllConversationsInactive on the restored GRDB.
  10. revokeStaleInstallations — builds a second throwaway
     client, calls XMTPInstallationRevoker. Non-fatal
     (revocation can fail without blocking the restore).
  11. SessionManager.resumeAfterRestore() — clears flags,
     next messagingService() call lazy-rebuilds the session
     against the restored identity + DB.

Rollback harness
- Pre-commit: any throw before the DB swap + any throw from the
  swap itself rolls back the XMTP file stash, resumes the
  session, rethrows, sets state = .failed. Mirrors the discipline
  on the old louis/icloud-backup branch.
- Post-commit: GRDB + XMTP state are already in the restored
  shape; further failures surface via state (.archiveImportFailed)
  but DO NOT attempt to undo. The plan §"Rollback harness" is
  explicit: no half-imported state gets unwound, the user hit
  "Restore" and consented.

State preservation
- If archive import fails, the subsequent revocation pass doesn't
  overwrite the .archiveImportFailed state — captured in a local
  var, re-pinned at the end so the caller sees the final signal.

Prerequisite tweaks
- AppEnvironment.iCloudContainerIdentifier (added in CP3c).
- DeviceInfoProviding.deviceName (added in CP3c).
- RestoreManager.init drops the default-arg for
  currentSchemaGeneration — LegacyDataWipe.currentGeneration is
  internal, and default-args can't reference internal symbols
  from a public initializer. Callers pass the generation
  explicitly; CP5's settings wiring will hand it in.

Tests (5 cases, RestoreManagerTests, .serialized)
- Happy path: create bundle via BackupManager, restoreFromBackup
  completes with state == .completed, importArchive recorded
  once on the mock, throwaway clientBuilder called twice (import +
  revocation).
- Discovery: findAvailableBackup returns the sidecar matching
  the one BackupManager wrote; conversationCount and
  schemaGeneration propagate.
- schemaGeneration mismatch: a RestoreManager built with a
  deliberately-different generation throws RestoreError.
- Identity timeout: short-timeout RestoreManager with empty
  keychain throws .identityTimeout within ~20ms.
- Archive import failure is non-fatal: mock's importArchive
  throws, restore still completes, state ends in
  .archiveImportFailed.

Deferred to CP3e
- Persisted archive bytes at <sharedContainer>/pending-archive-import.bin
  + "Retry history import" affordance in Settings. The state
  machine already emits .archiveImportFailed; CP3e just wires
  the persistence + UI surface.

Plan: docs/plans/icloud-backup-single-inbox.md §"Restore flow (new)",
§"Throwaway XMTP client for archive import", and the
`archiveImportFailed is retryable, not terminal` subsection.

feat(backup): TerminalSessionError + DeviceReplacedError + stale-check

CP4 of PR 3. Session-machine machinery that surfaces "this device
has been replaced" as a terminal error so the UI reset-flow (CP5)
has a state to observe. No UI in this commit — the banner +
conversations-list wiring lands alongside the rest of the
observer-side work in CP5 to keep CP4 pure-ConvosCore.

TerminalSessionError (new)
- Marker protocol. Empty protocol requirement — presence of the
  conformance is the contract.
- DeviceReplacedError: TerminalSessionError, Equatable. The only
  conformer today; future terminal conditions (hypothetical
  BackupRestoreCorruptedError, etc.) opt in via a one-line
  conformance.

SessionStateMachine changes
- handleRetryFromError short-circuits when
  `currentState == .error(e)` and `e is TerminalSessionError`.
  Prevents the retry-counter burn + "retry-by-coincidence lands
  in `.ready`" edge case the architect flagged in the Rev 4
  plan. Without this short-circuit, a background iCloud
  Keychain refresh between foreground attempts could silently
  flip the session to `.ready` on a revoked installation,
  masking the reset banner.
- handleAuthorized gains a stale-installation probe right
  before `emitStateChange(.ready(result))`:
    try await client.inboxState(refreshFromNetwork: true)
    if !state.installations.map(\.id).contains(client.installationId) {
        throw DeviceReplacedError()
    }
  Network failures during the probe log and proceed — the
  next foreground cycle retries. A successful probe that
  confirms revocation throws DeviceReplacedError, which the
  state machine catches and transitions to `.error(…)`.

Why handleAuthorized (and not elsewhere)
- Only authoritative place to catch the revocation:
  authenticatingBackend → ready is the first moment we have a
  live client that can read inboxState. Foreground-retry flows
  through the same helper via handleStop + handleAuthorize, so
  a single probe site covers both entry points per the plan's
  §"Stale-device as a session state, not a sidecar".

Tests (3 cases, TerminalSessionErrorTests)
- DeviceReplacedError conforms to TerminalSessionError.
- DeviceReplacedError is Equatable (default struct conformance).
- A non-conforming error type does NOT match `is
  TerminalSessionError`.
- The handleRetryFromError short-circuit + handleAuthorized
  probe require a live XMTP client and are exercised via the
  Docker-backed integration tests (not re-tested here — the
  marker-protocol contract pinned by this suite is the unit
  of coverage CP4 owns).

Plan: docs/plans/icloud-backup-single-inbox.md
§"Terminal errors: handleRetryFromError must not retry them"
and §"Stale-device as a session state, not a sidecar".

feat(backup): StaleDeviceBanner + ConversationsView wiring

CP5a of PR 3. UI observer-side half of CP4 — surfaces the
TerminalSessionError state to the user when the sole XMTP
installation for this inbox has been revoked.

StaleDeviceBanner (new)
- Single-variant pill: "This device has been replaced" with a
  destructive "Reset device" action and a bordered "Learn more"
  link. Rev 4 collapsed the old partialStale / fullStale
  variants into this one.
- Pure SwiftUI view, no VM of its own — takes two closures and
  lets the caller wire them.
- Accessibility: combined children, explicit label + identifier.

ConversationsViewModel wiring
- Subscribes to SessionStateMachine via
  `session.messagingService().sessionStateManager.observeState(_:)`
  — the same ClosureStateObserver + StateObserverHandle pattern
  the rest of the app uses.
- Flips isDeviceReplaced on every state change; the setter short-
  circuits no-op transitions so SwiftUI doesn't churn.
- `resetDevice()` action wraps `session.deleteAllInboxes()` in a
  MainActor Task + logs on failure.

ConversationsView
- sidebarContent now wraps its existing layout in a VStack; when
  isDeviceReplaced is true, StaleDeviceBanner renders above the
  list (or the empty state). Uses the same learn.convos.org URL
  as the InactiveConversationBanner from PR #725 so the two
  banners share an external resource.

No new tests in this commit — the banner itself is a pure view,
the VM wiring is a standard observer bridge, and the state
transitions it reacts to are already covered by
TerminalSessionErrorTests (CP4) + the Docker-backed integration
tests that exercise real revocation flows.

Plan: docs/plans/icloud-backup-single-inbox.md
§"Stale-device as a session state, not a sidecar" — banner's
"Reset device" action = session.deleteAllInboxes() per the plan.

docs(adr): ADR 012 — iCloud Backup (single-inbox architecture)

Captures what landed across CPs 1–5a of PR 3. Canonical
reference for reviewers who don't want to read the 8-commit
diff or the (now-mostly-historical) rev 4 plan doc. Supersedes
the vault-era plan documents authored on louis/icloud-backup.

Covers:
- Bundle format (CVBD magic + version header, tar, outer AES-GCM
  on identity.databaseKey, inner per-bundle archiveKey, sidecar
  vs full metadata projection).
- Create flow (BackupManager: skip guards → GRDB snapshot →
  client.createArchive → metadata → pack → seal → atomic iCloud
  write → sidecar after bundle).
- Restore flow (RestoreManager: discovery → awaitIdentity →
  decrypt + untar → validate → pauseForRestore → XMTP stash →
  replaceDatabase → throwaway Client.importArchive → commit →
  markAllConversationsInactive → revoke → resumeAfterRestore;
  rollback harness for pre-commit failures).
- NSE coordination (RestoreInProgressFlag + NSFileCoordinator
  write barrier).
- Device-replaced detection (TerminalSessionError marker +
  handleAuthorized probe + handleRetryFromError short-circuit +
  StaleDeviceBanner observer).
- Security model table and the list of things the bundle
  deliberately doesn't carry (media, keychain material,
  non-archive MLS history).
- Explicit pointer to CP5 follow-ups that are out of scope for
  this PR: BackupRestoreSettingsView + VM, BackupDebugView,
  BackupScheduler + xcconfig, fresh-install restore prompt,
  archiveImportFailed retry UX.

No source code changes in this commit — pure documentation.

> [!NOTE]
> ### Add iCloud backup and restore with AES-256-GCM bundle format, metadata, and device-replaced detection
> - Adds `BackupManager` to create encrypted iCloud backup bundles: snapshots the GRDB database, generates an XMTP archive, writes signed metadata, packs everything into a custom tar-like format, and encrypts with AES-256-GCM using the identity's database key.
> - Adds `RestoreManager` to discover and restore bundles: decrypts and validates the bundle, replaces the live database (with rollback), imports the XMTP archive, revokes stale installations, and exposes typed `RestoreState` progress transitions.
> - Introduces `BackupBundleCrypto` (AES-GCM, 32-byte CSPRNG keys), `BackupBundle` (custom tar with path-traversal hardening), and `BackupBundleMetadata` (full + secret-free sidecar for discovery without decryption).
> - Adds `TerminalSessionError` protocol and `DeviceReplacedError`; `SessionStateMachine` now probes active installations on authorization and enters a terminal error state if the local installation has been revoked, surfacing a `StaleDeviceBanner` in the conversations list.
> - `SessionManager` gains `pauseForRestore`/`resumeAfterRestore` and returns a `RestoreInProgressError` placeholder service while a restore is active; the notification service extension suppresses pushes during restore.
> - Risk: `replaceDatabase(with:)` performs an in-place swap of the live `DatabasePool` contents under `NSFileCoordinator`; double-fault (copy failure + rollback failure) throws `DatabaseManagerError.rollbackFailed` and leaves the database in an indeterminate state.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 98f55bd. 23 files reviewed, 10 issues evaluated, 2 issues filtered, 5 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> <details>
> <summary>Convos/Conversations List/ConversationsViewModel.swift — 0 comments posted, 1 evaluated, 1 filtered</summary>
>
> - [line 37](https://github.com/xmtplabs/convos-ios/blob/98f55bd1166df51f159f8ee42dc9a42753286c15/Convos/Conversations List/ConversationsViewModel.swift#L37): After `session.deleteAllInboxes()` completes successfully, the `isDeviceReplaced` flag remains stale because `sessionStateHandle` (created in `observeSessionState()`) is observing the old, now-deleted `MessagingService`. Any subsequent calls to `session.messagingService()` return a newly-created service whose state manager is not being observed. The banner will continue to show "This device has been replaced" even though the reset completed successfully, leaving the user with no indication the operation worked and no clear path forward. <b>[ Out of scope ]</b>
> </details>
>
> <details>
> <summary>ConvosCore/Sources/ConvosCore/Backup/RestoreManager.swift — 2 comments posted, 5 evaluated, 1 filtered</summary>
>
> - [line 128](https://github.com/xmtplabs/convos-ios/blob/98f55bd1166df51f159f8ee42dc9a42753286c15/ConvosCore/Sources/ConvosCore/Backup/RestoreManager.swift#L128): `findAvailableBackup` searches for backups in directories keyed by `DeviceInfo.deviceIdentifier` — the **current** device's ID. In a cross-device restore scenario (restoring Device B from Device A's backup), the backup lives in `.../backups/<DeviceA-id>/...` but this function looks in `.../backups/<DeviceB-id>/...`. The function will always return `nil` for backups created on a different device, silently failing to discover them. <b>[ Out of scope ]</b>
> </details>
>
>
> </details>
>

### DIFF
--- a/Convos/Conversations List/ConversationsView.swift
+++ b/Convos/Conversations List/ConversationsView.swift
@@ -124,14 +124,27 @@ struct ConversationsView: View {
 
     @ViewBuilder
     private var sidebarContent: some View {
-        if viewModel.unpinnedConversations.isEmpty && viewModel.pinnedConversations.isEmpty && viewModel.activeFilter == .all && horizontalSizeClass == .compact {
-            emptyConversationsViewScrollable
-        } else if viewModel.isFilteredResultEmpty && viewModel.pinnedConversations.isEmpty && horizontalSizeClass == .compact {
-            ScrollView {
-                filteredEmptyStateView
+        VStack(spacing: 0) {
+            if viewModel.isDeviceReplaced {
+                StaleDeviceBanner(
+                    onResetDevice: { viewModel.resetDevice() },
+                    onLearnMore: {
+                        if let url = URL(string: "https://learn.convos.org/") {
+                            UIApplication.shared.open(url)
+                        }
+                    }
+                )
+                .padding(.top, 8)
             }
-        } else {
-            conversationsCollectionView
+            if viewModel.unpinnedConversations.isEmpty && viewModel.pinnedConversations.isEmpty && viewModel.activeFilter == .all && horizontalSizeClass == .compact {
+                emptyConversationsViewScrollable
+            } else if viewModel.isFilteredResultEmpty && viewModel.pinnedConversations.isEmpty && horizontalSizeClass == .compact {
+                ScrollView {
+                    filteredEmptyStateView
+                }
+            } else {
+                conversationsCollectionView
+            }
         }
     }
 
--- a/Convos/Conversations List/ConversationsViewModel.swift
+++ b/Convos/Conversations List/ConversationsViewModel.swift
@@ -13,6 +13,38 @@ final class ConversationsViewModel {
 
     private(set) var focusCoordinator: FocusCoordinator
 
+    // MARK: - Device Replaced
+
+    private func observeSessionState() {
+        let messagingService = session.messagingService()
+        let manager = messagingService.sessionStateManager
+        sessionStateHandle = manager.observeState { [weak self] state in
+            let replaced: Bool
+            if case let .error(error) = state, error is DeviceReplacedError {
+                replaced = true
+            } else {
+                replaced = false
+            }
+            Task { @MainActor [weak self] in
+                guard let self else { return }
+                if self.isDeviceReplaced != replaced {
+                    self.isDeviceReplaced = replaced
+                }
+            }
+        }
+    }
+
+    func resetDevice() {
+        Task { @MainActor [weak self] in
+            guard let self else { return }
+            do {
+                try await self.session.deleteAllInboxes()
+            } catch {
+                Log.error("resetDevice: deleteAllInboxes failed: \(error)")
+            }
+        }
+    }
+
     // MARK: - Selection State
 
     @ObservationIgnored
@@ -180,6 +212,16 @@ final class ConversationsViewModel {
     private var cancellables: Set<AnyCancellable> = .init()
     @ObservationIgnored
     private var leftConversationObserver: Any?
+    @ObservationIgnored
+    private var sessionStateHandle: StateObserverHandle?
+
+    /// True when `SessionStateMachine` has transitioned to
+    /// `.error(DeviceReplacedError)` — the sole XMTP installation
+    /// for this inbox was revoked, almost always because the user
+    /// restored on another device. Drives the `StaleDeviceBanner`
+    /// at the top of the conversations list; the banner's "Reset
+    /// device" action runs `session.deleteAllInboxes()`.
+    var isDeviceReplaced: Bool = false
 
     private var horizontalSizeClass: UserInterfaceSizeClass?
 
@@ -279,6 +321,7 @@ final class ConversationsViewModel {
     }
 
     private func observe() {
+        observeSessionState()
         leftConversationObserver = NotificationCenter.default
             .addObserver(forName: .leftConversationNotification, object: nil, queue: .main) { [weak self] notification in
                 guard let conversationId = notification.userInfo?["conversationId"] as? String else {
--- a/Convos/Conversations List/StaleDeviceBanner.swift
+++ b/Convos/Conversations List/StaleDeviceBanner.swift
@@ -1,0 +1,69 @@
+import ConvosCore
+import SwiftUI
+
+/// Top-of-list banner shown when `SessionStateMachine` surfaces a
+/// `DeviceReplacedError` — i.e., the sole XMTP installation for
+/// this inbox has been revoked, almost always because the user
+/// restored on another device.
+///
+/// Single variant: Rev 4 collapsed the old `partialStale` /
+/// `fullStale` distinction into a binary. "Reset device" is the
+/// only path out; `Learn more` links to the convos learn site.
+struct StaleDeviceBanner: View {
+    let onResetDevice: () -> Void
+    let onLearnMore: () -> Void
+
+    var body: some View {
+        let resetAction = onResetDevice
+        let learnMoreAction = onLearnMore
+        VStack(spacing: 10) {
+            HStack(spacing: 8) {
+                Image(systemName: "exclamationmark.triangle.fill")
+                    .foregroundStyle(.colorLava)
+                    .font(.callout)
+                Text("This device has been replaced")
+                    .font(.callout)
+                    .fontWeight(.semibold)
+                    .foregroundStyle(.colorTextPrimary)
+                    .multilineTextAlignment(.center)
+            }
+            Text("Your account has been restored on another device. Resetting will clear local data on this device and restart setup.")
+                .font(.system(size: 13))
+                .foregroundStyle(.colorTextSecondary)
+                .multilineTextAlignment(.center)
+                .fixedSize(horizontal: false, vertical: true)
+
+            HStack(spacing: 8) {
+                Button(action: learnMoreAction) {
+                    Text("Learn more")
+                        .font(.system(size: 14, weight: .medium))
+                }
+                .buttonStyle(.bordered)
+
+                Button(role: .destructive, action: resetAction) {
+                    Text("Reset device")
+                        .font(.system(size: 14, weight: .medium))
+                }
+                .buttonStyle(.bordered)
+                .tint(.colorLava)
+            }
+        }
+        .frame(maxWidth: .infinity)
+        .padding(16)
+        .background(Color.colorFillMinimal, in: RoundedRectangle(cornerRadius: 16))
+        .padding(.horizontal, 16)
+        .accessibilityElement(children: .combine)
+        .accessibilityLabel(
+            "This device has been replaced. Your account has been restored on another device. Resetting will clear local data on this device and restart setup. Learn more or reset device."
+        )
+        .accessibilityIdentifier("stale-device-banner")
+    }
+}
+
+#Preview {
+    StaleDeviceBanner(
+        onResetDevice: {},
+        onLearnMore: {}
+    )
+    .padding()
+}
--- a/ConvosCore/Sources/ConvosCore/AppEnvironment.swift
+++ b/ConvosCore/Sources/ConvosCore/AppEnvironment.swift
@@ -109,6 +109,22 @@ public enum AppEnvironment: Sendable {
         }
     }
 
+    /// iCloud container identifier for the backup bundle. Derived from
+    /// the app-group identifier so it tracks per-environment naming
+    /// (dev / prod / local) without a second source of truth. The
+    /// entitlement must declare this container for the ubiquity URL
+    /// lookup to succeed; in tests the value is a valid-looking
+    /// placeholder that will resolve to `nil` at runtime (no account).
+    public var iCloudContainerIdentifier: String {
+        switch self {
+        case .local(config: let config), .dev(config: let config), .production(config: let config):
+            let bundleId = config.appGroupIdentifier.replacingOccurrences(of: "group.", with: "")
+            return "iCloud.\(bundleId)"
+        case .tests:
+            return "iCloud.org.convos.ios-local"
+        }
+    }
+
     public var keychainAccessGroup: String {
         // Use the app group identifier with team prefix for keychain sharing
         // This matches $(AppIdentifierPrefix)$(APP_GROUP_IDENTIFIER) in entitlements
--- a/ConvosCore/Sources/ConvosCore/Backup/BackupBundle.swift
+++ b/ConvosCore/Sources/ConvosCore/Backup/BackupBundle.swift
@@ -1,0 +1,236 @@
+import Foundation
+
+/// Tar format + pack/unpack for the iCloud backup bundle.
+///
+/// Format:
+///
+///     [4-byte magic "CVBD"]
+///     [1-byte format version]
+///     [entries...]
+///
+/// Each entry:
+///
+///     [4-byte path length BE][path UTF8][8-byte file length BE][file data]
+///
+/// The magic + version header lets future bundle-format bumps
+/// discriminate cleanly rather than decoding past-incompatible data.
+/// The entry format is a straight port of the old single-file-at-a-time
+/// layout; hardened against path-traversal on unpack via both a
+/// standardized-path check and a symlink-resolved re-check after
+/// `createDirectory`.
+public enum BackupBundle {
+    enum BundleError: Error, LocalizedError {
+        case directoryCreationFailed(String)
+        case packagingFailed(String)
+        case unpackingFailed(String)
+        case missingComponent(String)
+        case invalidMagic(expected: String, got: String)
+        case unsupportedVersion(got: UInt8, supported: UInt8)
+
+        var errorDescription: String? {
+            switch self {
+            case let .directoryCreationFailed(reason):
+                return "Failed to create backup directory: \(reason)"
+            case let .packagingFailed(reason):
+                return "Failed to package backup bundle: \(reason)"
+            case let .unpackingFailed(reason):
+                return "Failed to unpack backup bundle: \(reason)"
+            case let .missingComponent(name):
+                return "Missing backup component: \(name)"
+            case let .invalidMagic(expected, got):
+                return "Not a Convos backup bundle (expected magic \(expected), got \(got))"
+            case let .unsupportedVersion(got, supported):
+                return "Backup bundle format v\(got) is not supported (this build supports v\(supported))"
+            }
+        }
+    }
+
+    /// ASCII "CVBD" — Convos Backup Data.
+    static let magic: [UInt8] = [0x43, 0x56, 0x42, 0x44]
+
+    /// Current on-disk format version. Bump when a backwards-incompatible
+    /// structural change lands (e.g. media blobs, compressed entries).
+    static let currentFormatVersion: UInt8 = 1
+
+    public enum Component {
+        public static let database: String = "convos-single-inbox.sqlite"
+        public static let xmtpArchive: String = "xmtp-archive.bin"
+        public static let metadata: String = "metadata.json"
+    }
+
+    // MARK: - Staging
+
+    public static func createStagingDirectory() throws -> URL {
+        let tempDir = FileManager.default.temporaryDirectory
+            .appendingPathComponent("convos-backup-\(UUID().uuidString)", isDirectory: true)
+        do {
+            try FileManager.default.createDirectory(
+                at: tempDir,
+                withIntermediateDirectories: true
+            )
+        } catch {
+            throw BundleError.directoryCreationFailed(error.localizedDescription)
+        }
+        return tempDir
+    }
+
+    public static func cleanup(directory: URL) {
+        try? FileManager.default.removeItem(at: directory)
+    }
+
+    public static func databasePath(in directory: URL) -> URL {
+        directory.appendingPathComponent(Component.database)
+    }
+
+    public static func xmtpArchivePath(in directory: URL) -> URL {
+        directory.appendingPathComponent(Component.xmtpArchive)
+    }
+
+    // MARK: - Pack / unpack
+
+    public static func pack(directory: URL, encryptionKey: Data) throws -> Data {
+        let tarData = try tarDirectory(directory)
+        return try BackupBundleCrypto.encrypt(data: tarData, key: encryptionKey)
+    }
+
+    public static func unpack(data: Data, encryptionKey: Data, to directory: URL) throws {
+        let tarData = try BackupBundleCrypto.decrypt(data: data, key: encryptionKey)
+        try untarData(tarData, to: directory)
+    }
+
+    // MARK: - Tar (internal)
+
+    static func tarDirectory(_ directory: URL) throws -> Data {
+        var archive = Data()
+        archive.append(contentsOf: magic)
+        archive.append(currentFormatVersion)
+
+        let fileManager = FileManager.default
+        let resolvedDirPath = resolvedPath(directory)
+        let resolvedDirURL = URL(fileURLWithPath: resolvedDirPath, isDirectory: true)
+        guard let enumerator = fileManager.enumerator(
+            at: resolvedDirURL,
+            includingPropertiesForKeys: [.isRegularFileKey],
+            options: [.skipsHiddenFiles]
+        ) else {
+            throw BundleError.packagingFailed("failed to enumerate directory")
+        }
+
+        for case let fileURL as URL in enumerator {
+            let resourceValues = try fileURL.resourceValues(forKeys: [.isRegularFileKey])
+            guard resourceValues.isRegularFile == true else { continue }
+
+            let resolvedFilePath = resolvedPath(fileURL)
+            guard resolvedFilePath.hasPrefix(resolvedDirPath + "/") else {
+                throw BundleError.packagingFailed("file outside backup directory: \(fileURL.path)")
+            }
+            let relativePath = String(resolvedFilePath.dropFirst(resolvedDirPath.count + 1))
+            let pathData = Data(relativePath.utf8)
+            let fileData = try Data(contentsOf: fileURL)
+
+            var pathLength = UInt32(pathData.count).bigEndian
+            var fileLength = UInt64(fileData.count).bigEndian
+            archive.append(Data(bytes: &pathLength, count: 4))
+            archive.append(pathData)
+            archive.append(Data(bytes: &fileLength, count: 8))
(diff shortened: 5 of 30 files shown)


yewreeka:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/736?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#736** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/736?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/736?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#725** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=736).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #840, opened 2026-05-18
IAP Subscription and StoreKit Wiring

lourou:
Outcome-anchored paywall copy + low-balance banner in convo

Reworks paywall copy to drop model-tier and slow-mode bullets per
team feedback, leaning instead into outcome examples ("Plan ~5 trips",
"Run a daily agent for a month") under an "Enough credits each month
to:" header. Renames "assistants" to "agents" to match the team's
language. Flattens the bullet list to one VStack with uniform 8pt
spacing for an even rhythm, and tightens the paywall section/card
spacing to 24/12pt.

Stabilises the Subscribe button height by keeping the text label as
the layout-determining content and overlaying the progress spinner
during purchase, so the button no longer resizes on tap.

Adds an account-level LowBalanceBanner that pins above the messages
page in any conversation when the credit balance is low or depleted,
with an Upgrade CTA that presents the existing paywall. Gated to
non-production builds, same as the HOME credits pill.

Settings → Subscription row + detail screen

Adds the first user-facing entry point to the paywall outside the
debug menu. A new "Subscription" row in AppSettingsView surfaces the
current credit balance inline and navigates to a SubscriptionSettings
detail screen showing plan name, status, balance/grant, refresh date,
and Subscribe / Change plan / Manage in App Store actions.

The screen reads from the existing mock services so the Debug menu's
state-preset picker continues to drive every credits surface in lock-
step. Gated to non-production builds, matching the HOME credits pill
and the low-balance conversation banner.

Out-of-credits state on agent contact

Adds an out-of-credits section to the top of the agent member detail
screen when the account-level credit balance is depleted, with an
Upgrade button that presents the paywall. Scoped to agent members
only (non-agent contacts don't consume credits), gated to non-
production, and driven by the same mock credits publisher as the
HOME pill, the conversation low-balance banner, and the settings
subscription row.

Matches the rightmost CONTACT tile in Saul's credits sketch. Top up
is intentionally deferred ("fut" in the sketch); the per-agent usage
chart is a separate piece.

Make paywall bullets read as examples, not a contract

Initial outcome bullets ("Plan ~5 trips / Run a daily agent for a
month") landed as a promise rather than an illustration — testers
read them as "I pay to plan exactly 5 trips and run one daily agent
for a month". Reframes the bullets as explicit examples:

- Section intro becomes "For example, you could:" instead of
  "Enough credits each month to:" so the illustrative intent is
  baked into the lead sentence.
- A small italic caption below the bullets reads "Examples —
  actual usage varies by task." to defuse any remaining
  expectation that the bullets are a guarantee.
- Expanded to three outcome examples per tier (Draft ~100 emails
  for Builder; Plan a wedding for Pro) so the list reads as a
  sampler of possibilities, not an exhaustive entitlement.
- Outcome examples and tier features (Priority support on Pro)
  are now rendered as two visually distinct groups under separate
  intro lines ("For example..." vs "Includes:"), so the disclaimer
  applies cleanly only to the examples group.

Real StoreKit SubscriptionService wired behind a factory

Introduces a real StoreKitSubscriptionService that conforms to the
existing SubscriptionServiceProtocol, alongside the existing
MockSubscriptionService. Callers go through a new SubscriptionServices
factory; a FeatureFlags-style toggle picks mock vs real at runtime.
Non-production builds default to mocks so the debug state-preset
picker keeps working for design dogfood; flip the Debug → Use real
StoreKit toggle to point the paywall at real StoreKit (sandbox in
the sim via the new Convos.storekit configuration file).

The StoreKit service loads products via Product.products(for:),
runs purchase(options: [.appAccountToken(...)]), listens to
Transaction.updates, and exposes the current subscription via
the same publisher the UI already binds to. Receipt verification
against the backend is staged but skipped — the round trip to
POST /v2/subscriptions/me/verify is the obvious next step once
that backend route lands; the StoreKit-local Transaction state
still drives the UI in the meantime.

A Convos.storekit configuration file is added at the project root
and referenced from Convos (Dev).xcscheme so the sim can fetch the
four SKUs (Builder/Pro × Monthly/Annual) without App Store Connect
being fully provisioned. Product IDs are centralised in a new
SubscriptionProductIDs enum in ConvosCore so the mock catalogue,
the real StoreKit fetch, and the .storekit file all reference the
same strings.

All existing paywall callsites (LowBalanceBanner, contact
out-of-credits section, Settings subscription detail, Debug "Show
Paywall") now construct PaywallViewModel against
SubscriptionServices.shared instead of hardcoding the mock, so
flipping the toggle propagates everywhere without further changes.

NUX paywall step in onboarding

Adds a one-time paywall step to the conversation-onboarding state
machine between profile setup and the notifications prompt. The
existing PaywallView gets an optional onSkip closure: when set, the
close X is hidden and a "Start with a 7-day free trial" button
appears below the legal section, forcing the user to subscribe or
claim the trial.

ConversationOnboardingCoordinator gains a presentingPaywall state, a
shouldShowNUXPaywall gate (production-disabled, one-shot via
UserDefaults), and a userDidCompleteNUXPaywall() transition that the
sheet calls on either subscribe-success or skip-to-trial. The
PaywallViewModel exposes an onPurchaseSucceeded callback that fires
inside purchase() after the subscription service resolves, so the
coordinator advances without depending on publisher-side-effect
ordering.

ConversationOnboardingView binds a non-interactively-dismissible
sheet to the new state. The skip path sets MockCreditsService and
MockSubscriptionService to .trialActive as a placeholder for the
real POST /v2/credits/me/redeem-trial endpoint — easy swap once the
backend route lands.

Use system Manage Subscriptions sheet in Settings

Swaps the external "Manage in App Store" link in SubscriptionSettings
for StoreKit's native .manageSubscriptionsSheet modifier. Keeps the
user in-app, renders the system-provided UI for cancel / change-plan
/ pause, and works against both the local .storekit configuration in
the sim and real App Store Connect subscriptions in production.

Removes the obsolete openURL("https://apps.apple.com/account/subscriptions")
helper and the @Environment(\.openURL) dependency from the view.

Handle pending and unverified purchases distinctly

StoreKit Purchase outcomes that weren't outright failures were being
mapped to a generic "Purchase failed" error: Ask-to-Buy / SCA
challenges (which the user shouldn't perceive as an error) and Apple's
.unverified verification result (which has a different recovery hint
than a transient network failure).

Adds two new SubscriptionServiceError cases — purchasePending and
purchaseUnverified — and threads them through StoreKitSubscriptionService.
The mock service is unaffected since it never throws either case.

PaywallViewModel switches its alert state from a single
isShowingError/errorMessage pair to a titled alertTitle/alertMessage/
isShowingAlert triplet so each outcome can carry its own title:

  - purchasePending → "Awaiting approval" with a calm explainer
  - purchaseUnverified → "Couldn't verify purchase" + retry/restore hint
  - other failures → "Something went wrong" (unchanged behavior)
  - cancelled → still silent

The Transaction.updates listener already handles a pending purchase
resolving to approved later — refreshFromEntitlements() picks it up
and the subscription becomes active without further user action.

Wire iOS to backend credits + subscription endpoints

Hooks the IAP credits surfaces up to the real convos-backend contract
from louis/iap-credits-backend (PR 215):

  GET /v2/accounts/me/credits     → CreditBalance
  GET /v2/accounts/me/subscription → UserSubscription | 204
  POST /v2/accounts/me/subscription/verify { jwsRepresentation, appAccountToken }
                                           → { subscription: UserSubscription }

iOS-side changes:

- ConvosAPIClient gains getCreditBalance / getSubscription /
  verifySubscription, with VerifySubscriptionRequest + Response DTOs
  in the ConvosAPI namespace. Both response shapes already match the
  existing CreditBalance and UserSubscription structs (ISO 8601 dates
  decoded via the shared decoder); the 204-no-content case on
  subscription decodes to nil.

- BackendCreditsService — real CreditsServiceProtocol backed by the
  /v2/accounts/me/credits endpoint, mirroring MockCreditsService as a
  process-wide publisher singleton. refresh() runs once on init and
  can be called again on foreground/push.

- StoreKitSubscriptionService now relays every verified Transaction
  (purchase + Transaction.updates listener) to
  /v2/accounts/me/subscription/verify. Failures are logged but never
  fail the local purchase — Transaction.currentEntitlements stays
  authoritative for UI until the backend round trip resolves.

- New CreditsServices factory mirrors SubscriptionServices: production
  always returns the backend service; non-production reads a
  UserDefaults toggle so design dogfood can still hit the canned mock
  state presets. The HOME pill, conversation banner, contact section,
  Settings detail, and AppSettingsView row all read through the
  factory.

- Debug menu surfaces a "Use real backend credits" toggle alongside
  "Use real StoreKit". The mock-state preset picker hides when either
  real toggle is on, since the presets only manipulate mock services.

- ConvosAPIClientFactory + its protocol are now public so consumers
  outside ConvosCore (the StoreKit service in the main app) can
  construct an API client without going through ConvosClient.

- MockAPIClient gains stub implementations of the three new methods
  (zero balance, nil subscription, throwing verify) so test fixtures
  in ConvosCoreTests keep compiling.

Trial-claim from NUX is still a mock-only path
(MockCreditsService.shared.setPreset(.trialActive)) — backend
POST /v2/credits/me/redeem-trial isn't part of PR 215.

Drop appAccountToken from verify body; stabilise the install UUID

Per Borja's review on convos-backend PR 215: posting the appAccountToken
alongside the JWS lets a caller claim a transaction signed by Apple
for someone else's purchase. Apple's signature only proves the JWS
authenticity, not that the caller is the buyer.

The backend now extracts appAccountToken (and originalTransactionId)
directly from the signed payload and binds appAccountToken → accountId
on the first /verify hit. Subsequent verifies are trusted via the
JWT-authenticated account; a 409 fires if someone tries to re-bind
an already-claimed originalTransactionId to a different account.

iOS side:

- VerifySubscriptionRequest no longer carries appAccountToken; the
  body is just `{ jwsRepresentation }`. ConvosAPIClient.verifySubscription
  loses its `appAccountToken` parameter end-to-end.

- StoreKitSubscriptionService now persists a single UUID in
  UserDefaults (storeKit.appAccountToken) and reuses it for every
  product.purchase() call. The backend binds this token to the
  account on first verify, so consistent reuse keeps the
  binding stable across purchases on the same install.

- sendToBackendVerify drops the appAccountToken/fallbackToken
  plumbing entirely — both the JWS-embedded token and the JWT
  identify the buyer on the backend.

- MockAPIClient's stub signature updated to match.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/840"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

> [!NOTE]
> ### Add IAP subscription and StoreKit purchase flow with credits tracking
> - Adds `StoreKitSubscriptionService` as the production-ready `SubscriptionServiceProtocol` implementation: fetches products, verifies purchases with a backend JWS relay, publishes subscription state, and handles transaction updates and restore.
> - Introduces `PaywallView`/`PaywallViewModel` for a full subscription purchase UI with tier cards, period picker, error alerts, and optional trial-skip support.
> - Adds `BackendCreditsService` backed by a new `getCreditBalance()` API call with 15-second TTL debouncing; credits are force-refreshed after subscription changes.
> - Wires credits and subscription refreshes into app lifecycle (`scenePhase` active), conversations list appearance, and conversation view appearance.
> - Integrates a NUX paywall step into `ConversationOnboardingCoordinator` after profile setup, and exposes a `LowBalanceBanner` and `CreditsBadge` in the conversation and conversations-list views.
> - Adds `SubscriptionServices` and `CreditsServices` facade enums that switch between real and mock services via a persisted developer toggle in non-production builds.
> - Risk: All subscription and credits UI is gated to non-production environments; production paths use real StoreKit and backend APIs with no fallback if the JWS verification endpoint is unavailable.
>
> 
> #### Changes since #840 opened
>
> - Refactored subscription state management to derive subscription snapshots directly from verified transactions instead of querying `Transaction.currentEntitlements` [e42f509]
> - Updated `PaywallViewModel.purchase` to set the selected period to match the purchased product's period before invoking the purchase success callback [e42f509]
> - Removed `PaywallViewModel.hasActiveSubscription` computed property [b9205f3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8ca9680.</sup>
> 
>
>


lourou:
* **#862** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/862?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#849** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/849?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#840** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=840).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

Convos/Conversation Detail/ConversationMemberView.swift
```diff
@@ -21,6 +24,19 @@ struct ConversationMemberView: View {
...
         .scrollContentBackground(.hidden)
         .background(.colorBackgroundRaisedSecondary)
+        .onReceive(CreditsServices.shared.balancePublisher) { newBalance in
+            creditsBalance = newBalance
+        }
+        .task {
+            // Refresh credits when the contact sheet appears so the
+            // "out of credits" section + upgrade CTA reflect current
+            // backend state. TTL-debounced inside the service.
+            await CreditsServices.shared.refresh()
```
yewreeka, line 34: should we cache current credits in GRDB so we can follow the 'repository' pattern and have a way of displaying the current credit balance without having `onReceive` and a refresh `task` on every UI component?

yewreeka (dismissed):
- Credits should flow through GRDB + a `CreditsRepository`

  - `StoreKitSubscriptionService` could be an actor instead of `@unchecked Sendable` + `NSLock`, `lastFetchedAt` becomes isolated state, publisher/value getters get `nonisolated` ?

  - Drop the `Hermes` mention in `Convos/Conversation Detail/ConversationView.swift` (the `NOTE(hermes-burn-loop)` block).

  - Memoize the in-flight `loadProducts()` Task instead of the `isLoadingProducts` Bool, the second caller should await the same fetch
  *`Convos/Subscription/PaywallViewModel.swift`.*
```
private var loadProductsTask: Task<Void, Error>?

  func loadProducts() async throws {
      if let task = loadProductsTask { return try await task.value }
      let task = Task { try await actuallyLoad() }
      loadProductsTask = task
      defer { loadProductsTask = nil }
      return try await task.value
  }
```
  That way the second caller awaits the same in-flight load instead of returning early without products

Convos/Conversation Detail/ConversationMemberView.swift
lourou: @yewreeka thanks for the careful read!

- Hermes + Actor + Task memoization: doing now, will push shortly
- CreditsRepository: agreed, opening a follow-up PR!

fbac (approved):

PR #857, opened 2026-05-21
Performance improvements when catching up on messages

yewreeka:
perf(stream): skip ConversationWriter.store for typing + read-receipt events

Group messages hit conversationWriter.store() before content-type guards
ran. Typing indicators and read receipts then early-returned, but the
full conversation save + metadata extract + member fetch + profile
prefetch had already fired. In the slow-startup window from
convos-logs-C9FA22F1, 154 read receipts in 33s produced 154 full
saves on one conversation, none of which mutated any column on the
conversation row (read receipts persist to DBConversationReadReceipt
independently; typing indicators don't write at all).

Move the conversationWriter.store() call below the typing-indicator
and read-receipt guards. Explode-settings and profile-message paths
still require a fresh row (they read imageEncryptionKey / mutate
expiresAt), so they stay downstream of the store call. messageWriter
keeps its dbConversation parameter unchanged.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

perf(writer): skip no-op conversation saves to avoid spurious observer fires

When the same-id branch in saveConversation produces an updatedConversation
byte-identical to the existing row, the write was running anyway. GRDB
observers then fired, ConversationsRepository republished, and the
SwiftUI conversations list re-rendered for no state change. During the
slow-startup window from convos-logs-C9FA22F1, 232/233 save attempts
in 33s landed in this branch, almost all of them as no-ops driven by
read-receipt fanout (now also addressed by the StreamProcessor fix in
2d74091e). The other two branches — match-by-tag and brand-new — still
write because they create or replace rows that aren't byte-identical
to anything pre-existing.

Add an equality check on updatedConversation == existingConversation
before the .save(db) call in the same-id branch. DBConversation is
Hashable with all-primitive members, so the synthesized == is sound.
Log the skip at debug for diagnostics.

firstTimeSeeingConversationExpired semantics are preserved: when the
rows compare equal, their expiresAt fields are also equal, so the
existing isExpired && (expiresAt != existing.expiresAt) computation
already evaluates to false.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

perf(attestation): memoize negative verdicts and quiet stale-attestation logs

verifyCached() ran the full keyset lookup + signature verify path on
every call, even when the inputs were unchanged. During the slow-startup
window from convos-logs-C9FA22F1, agent 87378614 had its expired
attestation re-evaluated 214 times in 33s — three log lines per call —
because nothing memoized the verdict.

A negative verdict on (inboxId, kid, attestationTimestamp) is sound to
cache permanently: staleness fails monotonically for a fixed timestamp,
signature verification is deterministic on its inputs, and any change
to those inputs (rotated attestation, rotated keyset kid) produces a
different cache key. Positive verdicts are not cached because they can
transition to stale as time advances.

Use OSAllocatedUnfairLock<Set<NegativeCacheKey>> (matches the
established pattern in ImageCache, MockExplodeGroupOperations,
MockConversationStateManager) to keep verifyCached synchronous —
callers reach it from inside GRDB write transactions where adding an
await would ripple through the stack.

Also downgrade two info logs to debug:
- AssistantAttestationVerifier: "timestamp too old" fires once per
  unique key now (already), and is diagnostic detail, not signal.
- Profile.swift: "agent X cached result" fires every call regardless
  of cache hit; downgrade reduces redundant noise without losing the
  ability to inspect verdicts during debugging.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

perf(image-cache): persist URL tracker via per-entry sidecars

URLTracker was an in-memory-only [String: URL] map. Every cold launch
started empty, so peek() returned (changed: true, oldURL: nil) for every
identifier — which forced loadImage's "URL changed → fetch from network
first" branch and made every avatar do a network round trip on launch
even when the disk cache had a perfectly valid image. That's the
"images missing for a second on launch" regression.

Persist the map as one tiny .url file per identifier
(<sha256(identifier)>.url) co-located with the cached image in the disk
cache directory. URLTracker reads sidecars lazily on first peek/track
for an identifier (steady-state stays pure in-memory), writes them
atomically on track, and removes them on remove. removeImageFromDisk
also cleans up the .url file so cache eviction and sidecar lifetime
stay in lockstep.

This change touches URL tracker semantics only — loadImage's branch
selection still flows through peek(). After this lands, cold launches
on previously-cached identifiers report (changed: false) instead of
(changed: true), but the cache-first reorganization that turns that
into a UX win is the next commit. The two commits compose: this one
gives loadImage truthful information, the next one teaches it to use
that information without blocking on the network.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

perf(image-cache): cache-first loadImage with background refresh on URL change

loadImage's encrypted and unencrypted paths both started with a "URL
changed → fetch from network first, fall back to cache on failure"
branch. Combined with URLTracker's "missing entry == changed: true"
default, every cold-launch image lookup blocked on a network round trip
even when the disk cache had a perfectly valid image — that's the
"images missing for a second on launch" regression the user reported.

Flip both paths to cache-first:

  1. Memory hit → return immediately, refresh in background if URL drifted
  2. Disk hit → promote to memory, return, refresh in background if URL drifted
  3. Cold (no cached image at all) → fetch inline; this is the only path
     that blocks on the network

Combined with the sidecar-backed URLTracker (previous commit), cold
launches on previously-cached identifiers report `changed: false` and
skip the background refresh entirely — disk cache serves instantly with
zero network traffic. A real URL change (user updated their avatar)
still shows the stale image immediately and swaps in the fresh one when
the fire-and-forget refresh lands; the existing cacheUpdates observer
propagation handles the swap without any caller changes.

Inline fetch helpers handle their own deduplication (loadingTasksLock
keys fetchImageFromNetwork by URL; EncryptedImageLoader is idempotent
on the same encrypted ref), so multiple simultaneous background
refreshes for the same identifier coalesce safely.

Drops the previous "track-on-cache-hit when peek.changed is false"
re-tracking dance — the URL is already correctly tracked via the
sidecar, so re-tracking would be a no-op write.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

perf(prefetcher): make EncryptedImagePrefetcher a process-wide singleton

ConversationWriter.prefetchEncryptedImages instantiated a fresh
EncryptedImagePrefetcher per call. The prefetcher's actor-isolated
inflightFetches table is per-instance, so concurrent _store() calls
across N conversations each created their own prefetcher and each
checked ImageCacheContainer in parallel — all seeing "not cached" and
all firing fetches for the same inboxId. In the slow-startup window
from convos-logs-C9FA22F1, the user's own inbox got fetched 25 times
in 33 seconds because nothing coordinated across instances.

Add `EncryptedImagePrefetcher.shared` and route the writer through it.
Actor isolation now serializes `prefetchWithRetry`'s in-flight check
across all callers, so a second caller for an in-progress inboxId
awaits the existing task instead of duplicating the fetch. The
loader-injecting init stays for tests.

Cleanup runs unconditionally on completion (inflightFetches.removeValue
inside the `await task.value` lifecycle, which fires regardless of
fetch success or task cancellation since the task type is
Task<Void, Never>) so the dictionary doesn't grow.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

perf(profile): cache profile avatars by inboxId instead of per-conversation

Profile.imageCacheIdentifier returned "\(inboxId)@\(conversationId)",
so the same user's avatar was stored under a different cache key in
every conversation they appeared in. A user in 9 conversations meant
9 cache entries (and 9 fetches on first render across them), even
though the rendered avatar is logically one image per user.

Until a centralized profile-photo store exists (deferred — the
encrypted bytes are still per-conversation because each conversation
has its own imageEncryptionKey/salt/nonce/URL), key the in-memory and
on-disk cache by `inboxId` so a successful fetch from any conversation
satisfies every other conversation's render of the same user.

Cumulative effect with the singleton prefetcher (b5039dc0): same-user
fetches across N conversations collapse to one fetch + one cache write
+ N-1 cache hits, instead of N parallel fetches that race the
ImageCacheContainer cache check.

Existing per-conversation disk-cache entries become orphaned and will
be reclaimed by the cache eviction (capped at 500MB) over time. The
display name and metadata remain per-conversation; only the avatar
bytes are shared across conversations.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/857"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781927667&installation_id=128169237&pr_number=857&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F857&signature=98e689aafb5fae062b850fb3c25ece08b967026c188a7e4699a104923638ba35"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

> [!NOTE]
> ### Add batch catch-up to drain message backlog before streams resume on foreground and cold start
> - Introduces [`BatchCatchUp`](https://github.com/xmtplabs/convos-ios/pull/857/files#diff-397f4cd322efa095522345a526e59f4fc6b49ae617af5c733ae6a8c73f1e1dd0) which prepares conversations and messages concurrently, persists them in a single DB transaction, then applies reactions/read-receipts post-commit and dispatches side effects asynchronously.
> - [`SessionStateMachine`](https://github.com/xmtplabs/convos-ios/pull/857/files#diff-88a3691375b11b83b3fa12aab416b078bae5cc77821d7984f2c212efe87dd33e) runs the batch catch-up on both cold start and foreground resume using a `UserDefaults`-persisted cursor, advancing the cursor only after a successful run.
> - [`ConversationWriter`](https://github.com/xmtplabs/convos-ios/pull/857/files#diff-113bfd0c28b1a3b561f19e4574a0937082fe5c92a262e57573eee7728940a726) and [`IncomingMessageWriter`](https://github.com/xmtplabs/convos-ios/pull/857/files#diff-214371d5cccf7b16f4408fd58c74733a7d61314a74ebb428d08a97e20309c777) are refactored into `prepare`/`persist` steps so work can be parallelized before entering a write transaction.
> - [`ImageCache`](https://github.com/xmtplabs/convos-ios/pull/857/files#diff-445c3a2d9023d5d29469046555d79fa387c40b98bfcf772c84cdf1679fc38b91) now persists URL-change state across launches via on-disk `.url` sidecars, returning cached images immediately and refreshing in the background on URL drift instead of blocking on network.
> - Profile avatar cache keys change from per-conversation (`inboxId@conversationId`) to per-user (`inboxId`), so a fetched avatar is reused across all conversations for the same user.
> - Behavioral Change: `StreamProcessor` now skips saving the conversation row for typing indicators and read receipts, reducing DB writes during catch-up bursts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 694a2c3.</sup>
> 
>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/Crypto/AssistantAttestationVerifier.swift
+++ b/ConvosCore/Sources/ConvosCore/Crypto/AssistantAttestationVerifier.swift
@@ -61,7 +61,9 @@ public enum AssistantAttestationVerifier {
             return .unverified
         }
         guard let resolved = keyset.cachedResolveKey(for: kid) else {
-            negativeCache.withLock { _ = $0.insert(cacheKey) }
+            // Cache miss, not a verification failure. Don't poison the
+            // negative cache — a later `resolveKey()` may populate the
+            // keyset, and we want the next call to retry verification.
             return .unverified
         }
         let valid = verifySignature(
--- a/ConvosCore/Sources/ConvosCore/Crypto/AssistantAttestationVerifier.swift
+++ b/ConvosCore/Sources/ConvosCore/Crypto/AssistantAttestationVerifier.swift
@@ -1,7 +1,25 @@
 import CryptoKit
 import Foundation
+import os
 
 public enum AssistantAttestationVerifier {
+    /// Memoizes negative verdicts keyed by inputs that uniquely determine
+    /// the result. Staleness fails monotonically (age only grows for a
+    /// fixed `attestationTimestamp`) and signature verification is
+    /// deterministic, so once a `(inboxId, kid, attestationTimestamp)`
+    /// triple is known-unverified it stays unverified. When the agent
+    /// rotates to a new attestation the timestamp changes and the cache
+    /// key changes with it, so no false negatives. Positive verdicts are
+    /// not cached because they can transition to stale as time advances.
+    private struct NegativeCacheKey: Hashable {
+        let inboxId: String
+        let kid: String
+        let attestationTimestamp: String
+    }
+
+    private static let negativeCache: OSAllocatedUnfairLock<Set<NegativeCacheKey>>
+        = OSAllocatedUnfairLock(initialState: [])
+
     public static func verify(
         inboxId: String,
         attestation: String,
@@ -34,7 +52,16 @@ public enum AssistantAttestationVerifier {
         referenceDate: Date = Date(),
         maxAge: TimeInterval = 86400
     ) -> AgentVerification {
+        let cacheKey = NegativeCacheKey(
+            inboxId: inboxId,
+            kid: kid,
+            attestationTimestamp: attestationTimestamp
+        )
+        if negativeCache.withLock({ $0.contains(cacheKey) }) {
+            return .unverified
+        }
         guard let resolved = keyset.cachedResolveKey(for: kid) else {
+            negativeCache.withLock { _ = $0.insert(cacheKey) }
             return .unverified
         }
         let valid = verifySignature(
@@ -45,7 +72,11 @@ public enum AssistantAttestationVerifier {
             referenceDate: referenceDate,
             maxAge: maxAge
         )
-        return valid ? .verified(resolved.issuer) : .unverified
+        if !valid {
+            negativeCache.withLock { _ = $0.insert(cacheKey) }
+            return .unverified
+        }
+        return .verified(resolved.issuer)
     }
 
     private static func verifySignature(
@@ -68,7 +99,7 @@ public enum AssistantAttestationVerifier {
 
         let age = abs(referenceDate.timeIntervalSince(timestampDate))
         if age > maxAge {
-            Log.info("[Attestation] timestamp too old for \(inboxId.prefix(8)): age=\(Int(age))s, max=\(Int(maxAge))s")
+            Log.debug("[Attestation] timestamp too old for \(inboxId.prefix(8)): age=\(Int(age))s, max=\(Int(maxAge))s")
             return false
         }
 
--- a/ConvosCore/Sources/ConvosCore/Crypto/EncryptedImagePrefetcher.swift
+++ b/ConvosCore/Sources/ConvosCore/Crypto/EncryptedImagePrefetcher.swift
@@ -8,6 +8,16 @@ protocol EncryptedImagePrefetcherProtocol: Sendable {
 }
 
 actor EncryptedImagePrefetcher: EncryptedImagePrefetcherProtocol {
+    /// Process-wide singleton so `inflightFetches` dedupes across every
+    /// concurrent `_store()` invocation. Without it, each call site
+    /// instantiated its own prefetcher and the per-instance in-flight
+    /// table couldn't catch the actual race: many conversations going
+    /// through `_store()` in parallel all checking ImageCacheContainer
+    /// at the same time, all seeing "not cached," and all firing fetches
+    /// for the same inboxId. The init that accepts a custom loader stays
+    /// for tests.
+    static let shared: EncryptedImagePrefetcher = .init()
+
     private static let maxConcurrentDownloads: Int = 4
     private static let maxRetryAttempts: Int = 2
     private static let retryDelaySeconds: UInt64 = 1
--- a/ConvosCore/Sources/ConvosCore/Extensions/ImageCacheableExtensions.swift
+++ b/ConvosCore/Sources/ConvosCore/Extensions/ImageCacheableExtensions.swift
@@ -1,12 +1,17 @@
 import Foundation
 
 extension Profile: ImageCacheable {
-    /// Conversation-scoped cache key. A member's display name, avatar, and
-    /// metadata are all per-conversation (`DBMemberProfile` is keyed on
-    /// `(conversationId, inboxId)`), so keying by bare `inboxId` would
-    /// collapse distinct per-conversation profiles into one entry.
+    /// User-scoped cache key. The avatar is logically one image per user
+    /// even though each conversation stores its own encrypted copy
+    /// (separate `imageEncryptionKey`, salt, nonce, URL per conversation —
+    /// because we have no centralized profile-photo store yet). Keying by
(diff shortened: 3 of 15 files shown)


yewreeka:
* **#859** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/859?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#858** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/858?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#857** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/857?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/857?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=857).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #858, opened 2026-05-21
Catch up on messages when app enters foreground

yewreeka:
Replaces stream-replay backlog drain with a single batched fetch on
foreground/cold-start, then lets streams take over for live events.
Plan-only; implementation is a follow-up PR.

Key moves:
- One `list(lastActivityAfterNs:)` + parallel per-conv `messages(afterNs:)`
  fetch on foreground, before SyncingManager resumes streams
- All conversations + messages written in one GRDB transaction → one
  observer fire, one SwiftUI re-render
- Refactor `_store` and `store` into prepare(async) + persist(sync, in db)
  so the batch path can fan out the network work in parallel and collapse
  the persist into a single transaction
- Stream redelivery after the batch is free thanks to the no-op save
  short-circuit from #857 and existing message PK dedup — no new gating
  needed

Stacked on jarod/stream-write-storm-fixes (PR #857); the per-event cost
cuts there are necessary background even after this plan lands, because
stream catch-up still runs on every reconnect within a session.

Research verdict: feasibility-green-with-composition. No libxmtp upstream
changes needed.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/858"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781966959&installation_id=128169237&pr_number=858&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F858&signature=2f54e3a0d54cedc6c17e561243b432801b604f111f9ba21129c98e1086308a72"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

> [!NOTE]
[...]

PR #859, opened 2026-05-21
refactor(writer): split ConversationWriter._store into prepare/persist

yewreeka:
refactor(writer): split ConversationWriter._store into prepare/persist

Today's _store interleaves async XMTP network work (sync, members, profile
metadata extract) with the GRDB write transaction. That's fine for a
single stream event but blocks the batch catch-up plan: we want N
parallel `prepare` calls (fan out the network work across conversations)
followed by one transaction that persists all of them (collapse N
observer fires into one).

Introduce `PreparedConversation` + two new internal methods on
ConversationWriter:

  prepare(conversation: XMTPiOS.Group, inboxId: String, ...) async throws -> PreparedConversation
      Network-bound. Runs conversation.sync, extractConversationMetadata,
      conversation.members / memberProfiles, and createDBConversation.
      Transaction-free. Safe to call concurrently for many conversations.

  persist(_ prepared: PreparedConversation, in db: Database) throws -> ConversationSaveResult
      Synchronous. Call from inside `databaseWriter.write { db in ... }`.
      Runs the same DB-write logic the old `saveConversationToDatabase`
      closure ran — creator member upsert, saveConversation (with the
      no-op diff short-circuit), local-state insert, member roster
      reconciliation, profile gap-fill.

_store now calls `prepare(...)` then `databaseWriter.write { try persist(...) }`
then continues with the same post-save async side effects as before
(contact-sync enqueue, invite-tag self-heal, prefetch, invite gen,
fetchAndStoreLatestMessages, lastMessage store, processProfileMessagesFromHistory).
Identical observable behavior on the stream path.

The old saveConversationToDatabase wrapper is removed — persist + the
inline databaseWriter.write call in _store cover its job. ConversationSaveResult
loses its `private` modifier so it can flow out of `persist()` to
non-private callers (the upcoming BatchCatchUp actor).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

refactor(writer): split IncomingMessageWriter.store into prepare/persist

Same shape as the previous ConversationWriter refactor — extract the
async, transaction-free work into `prepare(message:)` and the
in-transaction body into `persist(_:conversation:in:)`. This unblocks
the batch catch-up coordinator from running N parallel prepares and
collapsing the persists into a single GRDB transaction.

Scope is the regular-message path only. Reactions (add/remove) still
flow through `handleReactionAddition` / `handleReactionRemoval`, each
with its own small transaction. Backlogs are typically text-heavy with
a sprinkling of reactions, so batching the text path captures most of
the observer-fire reduction; reaction overhead per event was already
neutralized by the no-op save short-circuit and prefetch dedup from
#857.

Behavior preserved on the stream path: `store(message:for:)` still
routes reactions first, then runs prepare + a `databaseWriter.write {}`
wrapping `persist(...)`. The drop-on-unverified-sender path now returns
`nil` from persist (instead of nil from the closure body) — same outer
behavior, just one level of indirection moved.

`PreparedIncomingMessage` carries the source DecodedMessage, the
encoded content type (avoids re-throwing the same decode), and the
pre-built DBMessage representation. `dbRepresentation()` runs in
prepare so multiple messages can build their reps in parallel before
the transaction.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

feat(syncing): BatchCatchUp coordinator for foreground backlog drain

A new internal actor that drains the conversation + message backlog
since a cursor in a single GRDB transaction, replacing the per-event
stream-replay path on foreground/cold-launch. Not yet wired into the
foreground hook — that's the next commit.

Flow:

  1. `client.conversationsProvider.listGroups(lastActivityAfterNs:)`
     discovers conversations that had activity since `since` (typically
     the NSE's `lastWelcomeProcessed` cursor; nil on cold start = fetch
     everything).
  2. Parallel-per-conversation prepare: read local `MAX(message.dateNs)`,
     fetch the message backlog from XMTP, filter to regular messages,
     build a `PreparedConversation` + `[PreparedIncomingMessage]` via
     the writers' prepare phases. All async / transaction-free.
  3. ONE `databaseWriter.write` transaction persisting all conversations
     and messages — one observer fire, one SwiftUI re-render.
  4. Per-conversation side effects (prefetch, invite generation,
     contact-sync, profile-message history) fire off the foreground
     critical path in a detached background task after the transaction
     commits.

Reactions, typing indicators, read receipts, and profile messages are
filtered out of the batch — each has its own per-event handler that
activates as soon as the stream takes over after the batch returns.
Stream redelivery of events the batch already wrote is idempotent
thanks to: the no-op save short-circuit from #857, DBMessage primary-key
INSERT OR REPLACE, and the reaction handler's existence check.

Also adds `ConversationWriter.runPostPersistSideEffects(prepared:group:)`
— the side-effect tail of `_store` extracted so the batch coordinator
can run it per-conversation off the critical path. Identical to
`_store`'s tail except `oldImageURL` is unconditionally nil (the batch
doesn't track per-row old URLs from a bulk persist; passing nil makes
the prefetcher fetch unconditionally, which is the right default for
a freshly-synced conversation).

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

feat(syncing): run batch catch-up on foreground before streams resume

Wires the BatchCatchUp coordinator (commit 3) into the foreground
lifecycle hook. On every `handleEnterForeground`, SessionStateMachine
now drains the message + invite backlog in batched form before letting
SyncingManager restart the message/conversation streams.

Hook shape:

  // SessionStateMachine.handleEnterForeground
  try await result.client.reconnectLocalDatabase()
  let cursor = readLastCatchUpCursor(for: inboxId)     // same UserDefaults
  await syncingManager?.runBatchCatchUp(                 // key the NSE uses
      client: result.client,
      since: cursor
  )
  writeLastCatchUpCursor(Date(), for: inboxId)
  await startNetworkMonitoring()
  await syncingManager?.resume()

`runBatchCatchUp` on SyncingManager is nonisolated so it doesn't enter
the actor's isolation domain — the only state it touches is the
immutable `identityStore` and `databaseWriter`, both safely shareable.
This avoids strict-concurrency tripping on the non-Sendable
`AnyClientProvider` being threaded through to BatchCatchUp /
InviteJoinRequestsManager.

Internally `runBatchCatchUp` calls two helpers sequentially:

- runMessageBatch — constructs fresh ConversationWriter +
  IncomingMessageWriter (stateless wrappers) + a BatchCatchUp and
  invokes its `run()`. Wrapped in do/catch so a batch failure is
  best-effort; the stream resume that follows is the safety net.
- runInviteBatch — constructs a fresh InviteJoinRequestsManager and
  calls `processJoinRequestOutcomes(since:client:)`, the same batch
  primitive the NSE uses on push-driven wakes. Same cursor as the
  message batch.

BatchCatchUp itself is now a `final class @unchecked Sendable` rather
than an actor: it's only constructed and `run` once per foreground (no
concurrent invocations possible), and being a class lets it accept the
non-Sendable client param without forcing every caller to wrap it.

Cursor (`convos.pushNotifications.lastWelcomeProcessed.<inboxId>`) is
read at the start of the hook and written at the end of the message
batch. The same UserDefaults key the NSE writes — foreground and NSE
share the catch-up frontier without double-processing. A small
duplication of the read/write helpers with MessagingService+PushNotifications
is accepted for now; extracting into a shared CatchUpCursor utility is
a follow-up.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/859"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781972468&installation_id=128169237&pr_number=859&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F859&signature=39ed737b5eeacee92434e7a744e4be6e56bebabca2f3d9bd1b2d9404ef6e4223"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

> [!NOTE]
> ### Split `ConversationWriter._store` into prepare/persist steps to support batched backlog catch-up
> - Introduces a `BatchCatchUp` coordinator that prepares conversations and messages in parallel, persists them in a single database transaction, then runs side effects and supplemental messages (reactions, read receipts) afterward.
> - Splits `ConversationWriter` and `IncomingMessageWriter` into `prepare` (async, network-bound, no DB writes) and `persist` (synchronous, transaction-scoped) methods to enable the batched path.
> - Adds `SyncingManager.runBatchCatchUp` to drain message and invite backlogs sequentially, keyed off a per-inbox `Date` cursor stored in `UserDefaults`.
> - `SessionStateMachine` calls `runBatchCatchUp` on cold start and on foreground resume, before starting streams.
> - `ConversationWriter.ConversationSaveResult` is promoted from `private` to internal visibility so batch catch-up callers can read its fields.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e693b23.</sup>
> 
>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
+++ b/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
@@ -810,7 +810,16 @@ public actor SessionStateMachine: SessionStateManagerProtocol {
         let inboxId = client.inboxId
         let cursor = Self.readLastCatchUpCursor(for: inboxId)
         Log.info("[catchup] running batch since=\(cursor.map { "\($0)" } ?? "nil") for inbox=\(inboxId.prefix(8))")
-        await syncingManager?.runBatchCatchUp(client: client, since: cursor)
+        // Don't advance the cursor unless we actually invoked the batch.
+        // `syncingManager` is nil when `AuthorizeInboxOperation` is configured
+        // with `startsStreamingServices: false` — silently advancing the
+        // cursor in that case would mark missed activity as "processed" and
+        // permanently skip it on the next foreground.
+        guard let syncingManager else {
+            Log.warning("[catchup] syncingManager nil, skipping batch (cursor unchanged)")
+            return
+        }
+        await syncingManager.runBatchCatchUp(client: client, since: cursor)
         Self.writeLastCatchUpCursor(Date(), for: inboxId)
     }
 
--- a/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
+++ b/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
@@ -773,11 +773,36 @@ public actor SessionStateMachine: SessionStateManagerProtocol {
 
         try await result.client.reconnectLocalDatabase()
 
+        // Drain the backlog in one batched transaction before streams resume.
+        // Replaces N per-event writes / observer fires / SwiftUI re-renders
+        // with one each. The cursor is the same `lastWelcomeProcessed`
+        // UserDefaults value the NSE writes after each push-driven catch-up,
+        // so foreground and NSE share the catch-up frontier without
+        // double-processing. The stream resume that follows is the
+        // correctness safety net for anything the batch missed.
+        let inboxId = result.client.inboxId
+        let cursor = Self.readLastCatchUpCursor(for: inboxId)
+        await syncingManager?.runBatchCatchUp(client: result.client, since: cursor)
+        Self.writeLastCatchUpCursor(Date(), for: inboxId)
+
         await startNetworkMonitoring()
         await syncingManager?.resume()
 
         emitStateChange(.ready(result))
         Log.info("Inbox returned to ready state")
+    }
+
+    /// Persisted catch-up cursor shared with `MessagingService+PushNotifications`.
+    /// Same UserDefaults key — both foreground and NSE update it as they
+    /// drain backlog, so the two paths converge on the same frontier.
+    private static let catchUpCursorKeyPrefix: String = "convos.pushNotifications.lastWelcomeProcessed"
+
+    private static func readLastCatchUpCursor(for inboxId: String) -> Date? {
+        UserDefaults.standard.object(forKey: "\(catchUpCursorKeyPrefix).\(inboxId)") as? Date
+    }
+
+    private static func writeLastCatchUpCursor(_ date: Date?, for inboxId: String) {
+        UserDefaults.standard.set(date, forKey: "\(catchUpCursorKeyPrefix).\(inboxId)")
     }
 
     private func handleRetryFromError() async throws {
--- a/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
+++ b/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
@@ -636,6 +636,16 @@ public actor SessionStateMachine: SessionStateManagerProtocol {
     }
 
     private func handleAuthorized(result: InboxReadyResult) async throws {
+        // Drain the backlog in batched form before streams start. Same
+        // motivation as `handleEnterForeground` — on cold start the
+        // network may have a large backlog (NSE didn't run, or the app
+        // was killed for a long time) and per-event stream catch-up
+        // would N-times the writes / observer fires / SwiftUI renders.
+        // The cursor is the same `lastWelcomeProcessed` UserDefaults
+        // value the NSE writes, so even cold start respects whatever the
+        // NSE has already processed in the background.
+        await runBatchCatchUp(client: result.client)
+
         await syncingManager?.start(with: result.client, apiClient: result.apiClient)
         foregroundRetryCount = 0
         emitStateChange(.ready(result))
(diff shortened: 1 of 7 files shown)


yewreeka:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/859?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#859** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/859?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/859?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#858** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/858?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#857** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/857?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=859).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #892, opened 2026-05-28
Key feed visibility on conversation consent

yewreeka:
## Summary

Supersedes the closed #885. Replaces the stored quarantine flag + hourly `QuarantineSweeper` with a system where **conversation consent is the source of truth for feed visibility**. The main feed queries the `[.allowed]` consent scope; nothing else gates visibility.

This started (in #885) as a read-side SQL predicate (`visibleInFeedPredicate` = creator is self or a non-blocked contact), but that dropped the "I explicitly joined this conversation" signal — a joined conversation whose creator wasn't a contact stayed hidden. Keying on consent fixes that cleanly.

## How it works

- **Arrival** (`StreamProcessor.decideInboundConversation`): inbound consent is bumped `.unknown -> .allowed` only when the user requested to join (`hasOutgoingJoinRequest`) or the creator is already a non-blocked contact. Unsolicited strangers are delivered but left `.unknown` (out of the feed scope, hidden).
- **Promotion / demotion / restoration** (`ConversationConsentReconciler`, new): a `ValueObservation` on the `contact` table keeps each contact-created conversation's consent in sync with block state — non-blocked contact -> `.allowed`, blocked -> `.denied`. One convergent reconciler covers "stranger becomes a contact" (promote), "block" (demote), and "unblock" (restore), so `block`/`unblock` stay pure contact-row writes. Consent is flipped at the XMTP layer (survives re-sync) and mirrored to the DB; it skips the network call when XMTP already matches the target.
- **Cleanup**: the stale-stranger GC deletes conversations still `.unknown` past the 7-day TTL.
- **Removed**: `visibleInFeedPredicate`, the contact/inbox observation-region widening from the feed + count repos, and the dead `quarantinedAt`/`quarantineReleasedAt` columns (with a `dropConversationQuarantineFields` migration).

## Testing

- Full `ConvosCore` suite green: **992 tests / 142 suites passed** (Docker integration suites + the drop-column migration included).
- New `ConversationConsentReconcilerTests` (promote / demote / no-op / leave-non-contacts-alone).
- Rewritten stranger-visibility test for the consent model.
- Verified on-device: a conversation joined from another device (creator not a contact) now appears in the feed; previously hidden.

## Known follow-ups (non-blocking)

- **Stuck-`.unknown` edge**: a joined/self conversation that misses its arrival-flip isn't re-healed by the reconciler (which only joins contacts) and could be GC'd after 7 days. Rare; tracking.
- **Pre-existing (out of scope)**: "join a group you created -> auto-contact -> their future conversations are visible with no consent gate" lives in `ContactSyncCoordinator`, unchanged here. Worth a separate ticket.

## Migration

No data migration needed. #885 never shipped (PR closed, not merged), so the intermediate "flip all inbound to `.allowed`" behavior never ran on any release; production keeps strangers `.unknown`. The only schema change is dropping the two dead quarantine columns (safe at min iOS 26 / SQLite 3.35+).

> [!NOTE]
> ### Replace conversation quarantine with consent-based visibility and stale stranger garbage collection
> - Removes the quarantine model entirely: `quarantinedAt`/`quarantineReleasedAt` columns are dropped via a DB migration, removed from `DBConversation`, and all related persistence/filter logic is deleted.
> - Adds `ConversationConsentReconciler`, a background observer that detects conversations whose XMTP consent state disagrees with the creator's contact/block status and reconciles both the XMTP layer and the local DB.
> - Simplifies `InboundConversationFilter` to a stateless, no-arg struct: `.allowed` and `.unknown` conversations are persisted (deliver); only `.denied` is rejected. Contact-aware quarantine branching is removed.
> - Adds a stale-stranger GC in `SessionManager` that hard-deletes `.unknown`-consent conversations older than 7 days, running at launch and every hour; replaces the old `QuarantineSweeper` and `contactBlockingDidChange` observer.
> - Risk: conversations from unknown contacts are now persisted immediately (at `.unknown` consent) and become visible once consent is promoted to `.allowed`; the old quarantine holding area no longer exists.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bd3dc79.</sup>
> 
>
>


yewreeka:
Closing — rolling this into the agent-contacts refactor that builds on #854 now landed on dev.

PR #971, opened 2026-06-03
Add iCloud-synced keychain backup slot for identity recovery

yewreeka:
## What

The user's identity (secp256k1 private key) currently lives only in a device-local keychain slot, so a lost phone means a lost identity. This adds a second, iCloud-synced keychain slot that backs up the identity for recovery — while keeping device sync an explicit user action.

## Design

| | Primary slot | Synced backup slot |
|---|---|---|
| Service | `…KeychainIdentityStore.v3` (unchanged) | `…KeychainIdentityStore.v3-synced-backup` (new) |
| `kSecAttrSynchronizable` | `false` | `true` |
| Account | `convos-identity` (fixed) | the identity's `inboxId` (one item per identity) |
| Contents | full identity incl. database key | `KeychainIdentityBackup`: inboxId, clientId, private key + restore-display metadata (device name, backup date) |
| Read at runtime | yes (`load`/`loadSync`) | never — future recovery flow only |

- **Sync stays explicit**: nothing in the runtime read path touches the backup slot, so another device sharing the iCloud account never silently adopts the identity.
- **Per-identity accounts**: unpaired identities on devices sharing an iCloud account back up side by side instead of clobbering one fixed slot (covers both save and delete).
- **Displacement cleanup**: when a save overwrites a *different* inboxId (pairing over the fresh-install placeholder), the displaced identity's backup is removed rather than orphaned in iCloud.
- **Database key excluded** from the backup: it only decrypts this device's local DB, which never leaves the device — escrowing it would add exposure with no recovery value. A recovering device generates a fresh one.
- **Restore-display metadata**: each backup is stamped with the user-visible device name (same `DeviceInfo.deviceName` the pairing flow uses, lazily injected) and the write date, so the restore picker can show "Alice's iPhone, backed up June 3" instead of a bare inboxId. Both fields decode as optional so missing metadata can never make a backup unrecoverable.
- **App Clip never escrows**: the clip constructs the store with `syncedBackupEnabled: false`; its auto-registered identity is backed up only after the full app adopts it (authorize-time backfill).
- **Backfill on authorize** covers existing installs that registered before the backup slot existed.
- **`delete()` scopes** backup removal to the deleted identity's inboxId; if the primary is unreadable the backup is left as a logged orphan (documented trade-off — sweeping would destroy other devices' backups; `LegacyDataWipe` reclaims orphans on generation bumps).
- **`LegacyDataWipe`** sweeps the new service (`kSecAttrSynchronizableAny`) on generation bumps; note the cross-device blast radius is documented inline and is intentional for a total reset.

## CI fix

The `KeychainIdentityStore Tests` workflow pinned `OS=26.1`, which the macos-26 runner image no longer carries (job has been failing on infra since the image moved to 26.2+; last green run on dev was May 1). Dropped the OS pin so it matches any available iPhone 17.

## Review process

Four parallel review agents (security/key-material, Security-framework correctness + concurrency, lifecycle/integration edge cases, test adequacy) reviewed the initial implementation. All major findings were addressed: App Clip escrow gating, displaced-backup cleanup in the pairing flow, database-key exclusion, the synchronizable `SecItemUpdate` re-save test, and per-test keychain sweeps in the integration suite.

## Open product questions

- A displaced identity's backup is deleted (no zombie keys in iCloud). If we ever want displaced identities to remain recoverable, that's a deliberate product decision to make later.
- Generation bumps wipe all backups across all devices on the account (existing wipe semantics, now with bigger blast radius). Flagged inline; needs no action unless we bump generations casually.

## Testing

- `KeychainSyncConfigTests` (unit, mock): 18/18 — incl. displacement, re-save dedupe, metadata-optional decoding, JSON-level proof the database key is absent from the backup blob
- `KeychainIdentityStoreRealKeychainTests` (real keychain on simulator): 23/23 — incl. pinned sync/non-sync placement queries for both slots, two-identity coexistence, displacement, the `errSecDuplicateItem → SecItemUpdate` re-save path on the synchronizable item, device-name/date stamping, orphan edge, corrupt-blob skip, 3-item enumeration
- SwiftLint strict + SwiftFormat clean; full ConvosCore suite runs in CI

> [!NOTE]
> ### Add iCloud-synced keychain backup slot for identity recovery
> - Adds a `KeychainIdentityBackup` struct and an iCloud-synced secondary keychain slot (service `org.convos.ios.KeychainIdentityStore.v3-synced-backup`) that mirrors the primary identity on every `save()`.
> - Introduces `loadSyncedBackups()` and `backfillSyncedBackupIfNeeded()` on `KeychainIdentityStoreProtocol`; backfill is triggered during the `SessionStateMachine` authorization flow.
> - Backup metadata includes device name (stamped via a `deviceNameProvider` closure) and a timestamp; the App Clip disables synced backups via `syncedBackupEnabled: false`.
> - Legacy data wipe now sweeps the synced backup service so generation bumps propagate deletions across devices via iCloud.
> - Risk: the synced slot uses `kSecAttrSynchronizable = true`, meaning backup items can appear on other devices signed into the same iCloud account before the user explicitly initiates recovery.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e076f5.</sup>
> 
>

### DIFF
--- a/.github/workflows/keychain-identity-store-tests.yml
+++ b/.github/workflows/keychain-identity-store-tests.yml
@@ -63,7 +63,7 @@ jobs:
           xcodebuild test \
             -project Convos.xcodeproj \
             -scheme KeychainIdentityStoreTests \
-            -destination 'platform=iOS Simulator,name=iPhone 17,OS=26.1' \
+            -destination 'platform=iOS Simulator,name=iPhone 17' \
             -enableCodeCoverage YES \
             -skipPackagePluginValidation \
             -resultBundlePath TestResults.xcresult \
--- a/ConvosCore/Sources/ConvosCore/Auth/Keychain/KeychainIdentityStore.swift
+++ b/ConvosCore/Sources/ConvosCore/Auth/Keychain/KeychainIdentityStore.swift
@@ -79,6 +79,44 @@ public struct KeychainIdentity: Codable, Sendable {
     public let keys: KeychainIdentityKeys
 }
 
+/// Identity material mirrored into the iCloud-synced backup slot.
+/// Deliberately excludes the SQLCipher database key: it only decrypts
+/// this device's local database, which never leaves the device, so
+/// escrowing it in iCloud would add exposure without recovery value.
+/// A recovering device generates a fresh database key.
+public struct KeychainIdentityBackup: Codable, Sendable {
+    public let inboxId: String
+    public let clientId: String
+    public let privateKey: PrivateKey
+
+    private enum CodingKeys: String, CodingKey {
+        case inboxId
+        case clientId
+        case privateKeyData
+    }
+
+    init(identity: KeychainIdentity) {
+        inboxId = identity.inboxId
+        clientId = identity.clientId
+        privateKey = identity.keys.privateKey
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        inboxId = try container.decode(String.self, forKey: .inboxId)
+        clientId = try container.decode(String.self, forKey: .clientId)
+        let privateKeyData = try container.decode(Data.self, forKey: .privateKeyData)
+        privateKey = try PrivateKey(privateKeyData)
+    }
+
+    public func encode(to encoder: Encoder) throws {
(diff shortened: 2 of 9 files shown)


lourou (approved):
LGTM

PR #982, opened 2026-06-04
Fix message-list animations: receipts, sends, reactions, composer interplay

yewreeka:
Fixes four families of message-list animation jank. The history is squashed to one commit (the iterative exploration added and later removed scaffolding that made per-commit review misleading) and rebased onto dev to integrate with #987 and #998.

## Receipts: Sent -> Read morph

The status row's swap rendered as a one-frame hard cut: cell reconfigures evaluate the hosting view in a sizing pass before the render pass, which consumes value-based `.animation(value:)` triggers (verified with transaction probes). The row now renders from a seeded `readByMembers` mirror mutated inside `withAnimation` from `onChange` -- a transaction the two-pass update can't eat. The caption's layout space is reserved while a message is unpublished (it fades in at publish), so the status flip is height-neutral.

## Sends: grow below the fold, reveal with a scroll

A sent message appends to the bottom group's cell (one group = one cell), so a send is a height-growing reconfigure whose bottom-pinning compensation applied instantly -- a one-frame ~43pt jump, twice per send. In-cell animation cannot mask it: the hosting view reports the final height in a single atomic invalidation (verified across five masking strategies). Instead, the group renders from an `animatedGroup` mirror applied without animation in `onChange`: the sizing pass reports the old height, the growth lands below the fold invisibly, and an animated scroll reveals it. The layout skips immediate compensation for large out-of-band growth and notifies the controller (`onOutOfBandBottomGrowth`) so growth without a state update (async media) re-pins the same way. The re-pin is gated on a tracked `isPinnedToBottom` flag (updated on offset changes), so receipts/reactions/typing updates never yank a reading user to the bottom.

## Reactions: native batch path

Reactions render from the **live** group rather than the mirror, so their height change is measured in the sizing pass and lands inside the batch update, where UIKit animates surrounding cells and compensation natively -- the smooth pre-existing behavior validated again after #998. (Earlier revisions routed reactions through the mirror and rebuilt the animation with a per-frame squeeze; that scaffolding is gone.)

## Composer interplay

- Reacting from the context menu no longer bounces the list down/up: keyboard-driven inset updates stay suppressed while the menu is up, and the returning keyboard matches the preserved inset with zero motion; a deferred fallback settles if focus never returns.
- Sending a multi-line draft no longer dips the list below the bar: the composer collapse's inset change is deferred (`pendingComposerBottomInset`) and flushed at the start of the reveal scroll, where it applies clamp-free against the grown content. #987's `applyDeferredBottomInset` converts shrink-while-pinned flushes into the same deferral so its silent path can't reintroduce the clamp jump.
- `MessagesCollectionView` applies quiescent layout passes without inherited animations (explicit UIView animations like keyboard/rotation are exempt), and cells top-anchor their hosting content so transient cell/content height mismatches stay below the fold.

## Verification

Every fix was verified by frame-diffing and landmark-tracking 60fps simulator recordings (profiles in the review thread): sends are a monotonic ease ramp with no dip frame; the multi-line composer send shows zero downward displacement; the context-menu reaction flow measures +0px end to end; reactions match the dev/#998 motion profile. A four-agent adversarial review of an earlier revision surfaced and fixed two high-severity gates (re-pin tolerance, shrink compensation). Full `swift test --package-path ConvosCore` run green on the pre-rebase head; final suite run on this head reported separately.

> [!NOTE]
> ### Fix message-list animations for receipts, sends, reactions, and composer interplay
> - Introduces a mirrored `animatedGroup` state in [`MessagesGroupView`](https://github.com/xmtplabs/convos-ios/pull/982/files#diff-96fea19951f287381d9170a90ce9ac8ba6f9e59f9d477dd1c5be2bbc1b5cdbe2) so group-level layout changes apply after the sizing pass, preventing instant frame jumps while enabling scoped content animations.
> - Defers composer inset shrinks while pinned to bottom in [`MessagesViewController`](https://github.com/xmtplabs/convos-ios/pull/982/files#diff-1427b816e5e93d8bc7ea882f5eabd9671135c6f9b9ebc6b0a6f7116a2f8a37c3) to avoid clamp or drag artifacts; flushes deferred insets before scrolling to bottom.
> - Adds `onOutOfBandBottomGrowth` callback in [`MessagesCollectionLayout`](https://github.com/xmtplabs/convos-ios/pull/982/files#diff-4c22b0b8d30661ba1bf904fb3bfb896554201ff97fe825099213e2ed967066d7) so large self-sizing height increases outside batch updates trigger an animated scroll instead of an immediate offset jump.
> - Overrides `layoutSubviews` in [`MessagesCollectionView`](https://github.com/xmtplabs/convos-ios/pull/982/files#diff-cb9c0963c13db853c9bf20ed227f212d6e890a556de39835348635ec0dc58873) to suppress inherited animations during quiescent layout passes, preserving animation only during batch updates or explicit `UIView` animations.
> - Restores bottom inset after context menu dismissal with a delayed fallback when the keyboard does not reappear, avoiding a visible bounce.
> - Read receipt avatars now morph with a spring animation; status row space is reserved for unpublished messages to prevent layout pops on send.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2e06c39.</sup>
> 
>

### DIFF
--- a/Convos/Conversation Detail/Messages/Messages Collection Layout/MessagesCollectionLayout.swift
+++ b/Convos/Conversation Detail/Messages/Messages Collection Layout/MessagesCollectionLayout.swift
@@ -136,6 +136,34 @@ open class MessagesCollectionLayout: UICollectionViewLayout {
 
     var hasPinnedHeaderOrFooter: Bool = false
 
+    /// Invoked (coalesced, on the next runloop tick) after the layout skips
+    /// the bottom-pinning compensation for out-of-band self-sizing growth
+    /// (see `invalidationContext(forPreferredLayoutAttributes:withOriginalAttributes:)`).
+    /// The owner reveals the growth with an animated scroll when the list
+    /// was pinned to the bottom.
+    var onOutOfBandBottomGrowth: (() -> Void)?
+
+    private var hasPendingOutOfBandGrowthNotification: Bool = false
+
+    private func notifyOutOfBandBottomGrowth() {
+        guard !hasPendingOutOfBandGrowthNotification else { return }
+        hasPendingOutOfBandGrowthNotification = true
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            hasPendingOutOfBandGrowthNotification = false
+            onOutOfBandBottomGrowth?()
+        }
+    }
+
+    /// True while the layout is processing a batch update (between
+    /// `prepare(forCollectionViewUpdates:)` and the state switch after
+    /// `finalizeCollectionViewUpdates`). `MessagesCollectionView` uses this
+    /// to keep batch layout passes animated while applying all other passes
+    /// without inherited animations.
+    var isInBatchUpdates: Bool {
+        state == .afterUpdate
+    }
+
     // MARK: Constructors
 
     init(flipsHorizontallyInOppositeLayoutDirection: Bool = true) {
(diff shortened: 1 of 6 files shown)


yewreeka:
* **#982** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/982?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/982?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=982).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #986, opened 2026-06-04
Fix clean-install bootstrap conversations flashing as visible rows

yewreeka:
## Summary

On a clean install with zero user interaction, two "New Convo" rows appeared in the chats list: the prewarmed cache conversation flashed visible for ~14s before flipping back to hidden, and the inline agent builder's auto-created draft became a permanently visible row. Both were captured live on a factory-erased simulator (DB polling at 50ms + screen recording + app logs).

**Repro trace (factory-clean sim, no interaction):**

| time | event |
|---|---|
| 16:44:58 | Agent-builder auto-create publishes group `8b5670…`, stored `isUnused=0` -> visible forever |
| 16:45:38 | Prewarm group `7671ae…` echoed back by the stream, stored `isUnused=0` -> visible |
| 16:45:52 | Cache's own write flips `7671ae…` back to `isUnused=1` -> 14-second flash |

## Root cause

The store path defaults every new conversation row to visible (`isUnused: false`), and "hidden" was only enforced after the fact:

1. **Prewarm flash**: `UnusedConversationCache.runPreparation()` published the group on XMTP, then did more async work (`ensureInviteTag`, `ensureImageEncryptionKey`), and only then wrote the hidden DB row. In that window the conversation stream echoed the group back and `ConversationWriter` inserted it as a fresh visible row; the cache's later write flipped it back (the code even called this a "benign collision").
2. **Agent-builder phantom row**: `.newAgent` mode auto-creates its conversation on mount. With an empty cache (always true on first install, prewarm still in flight), `prepareNewConversation()` returns nil and `ConversationStateMachine.handleCreate` publishes a brand-new group that gets stored visible. The deferred-commit logic only protected the cache-claimed path; the auto-created row had no claimed id, so nothing ever hid or committed it.

## Fix

Make "hidden until committed" hold from the row's birth, for both creation paths:

- **`UnusedConversationCache`**: write the hidden row *before* `publish()` so the stream echo finds an existing `isUnused = true` row and preserves it (the by-id update path already preserves hidden state). The cache self-claims the id for the whole preparation window so `consumeUnusedConversationId` can't hand out a group that hasn't finished publishing. Rollback now also covers the pre-publish row (delete row; `leaveGroup()` only if publish succeeded).
- **`ConversationStateMachine.handleCreate(startsUnused:)`**: new flag threaded from `NewConversationViewModel` (`.newAgent` only) through `DraftConversationWriterProtocol.createConversation(startsUnused:)`. When set, the same hidden-row shape (shared `UnusedConversationCache.writeUnusedConversationRow`) is written between `prepareConversation()` and `publish()`.
- **Commit side**: `NewConversationViewModel.commitConversationVisibility()` promotes the row at the Make tap. Covers the cache-claimed row (id known at acquire) and the auto-created row (id adopted as `claimedConversationId` at `.ready`, registered with the cache claim set so no other flow can consume it while the builder is alive). A Make tap that lands before `.ready` queues the commit and flushes it at `.ready`, mirroring how message sends queue.
- **`ConversationWriter.saveConversation`**: the match-by-invite-tag branch now preserves `isUnused` like the match-by-id branch already did — a row only becomes visible through an explicit commit, never via a stream echo replacing it.

Abandoned hidden rows (app killed mid-builder) are claimable by the cache again on next launch (claims are in-memory), so they get recycled as the next prewarmed conversation instead of accumulating.

## Bonus effects

- `hasAnyUsedConversations()` (pairing destructive-warning gate) is now accurate on fresh installs: bootstrap rows stay `isUnused = true` until real engagement.
- The inline agent builder no longer unmounts/remounts on first install (its own phantom row was flipping the list out of the empty state).

## Verification

- Fresh-install run on a factory-erased simulator: conversation table shows both bootstrap rows `isUnused=1` from creation; chats list stays empty; no flash.
- Make-tap flow: typed a prompt, tapped Make — row flips `isUnused=0`, conversation appears in the list, sheet presents.
- App builds (`Convos (Dev)`); SwiftLint clean on changed files.

> [!NOTE]
> ### Fix bootstrap conversations flashing as visible rows during clean-install agent creation
> - New conversations created in `.newAgent` mode are now written as hidden rows (via `startsUnused`) before being published to the network, preventing them from appearing in the chats list prematurely.
> - Introduces deferred visibility coordination in [`NewConversationViewModel`](https://github.com/xmtplabs/convos-ios/pull/986/files#diff-bc0af06ca6587e88461a662416743a0ef38cc55d66fc169caad65a62fb3abe47): a `commitConversationVisibility()` method promotes hidden rows to visible, handling both cache-claimed and auto-created paths, including taps that arrive before `.ready`.
> - [`UnusedConversationCache`](https://github.com/xmtplabs/convos-ios/pull/986/files#diff-927068b0d884e130c12a37c911c53ee25bd89d9fe2c772600c66cd42edd25220) now writes hidden rows with a provisional invite tag before publish, so prewarmed conversations are never visible mid-preparation; stream echoes replacing hidden rows by invite tag no longer flip them visible.
> - [`ConversationWriter`](https://github.com/xmtplabs/convos-ios/pull/986/files#diff-113bfd0c28b1a3b561f19e4574a0937082fe5c92a262e57573eee7728940a726) preserves `isUnused == true` when upserting by invite tag to prevent network echoes from revealing hidden rows.
> - Risk: The ordering contract between `claimRegistrationTask` and `commitConversationVisibility()` is new; a race during dismissal before `.ready` is mitigated by clearing `pendingVisibilityCommit` on deletion, but error-path rollbacks in `UnusedConversationCache` are more complex than before.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 282b70e.</sup>
> 
>

### DIFF
--- a/Convos/Agent Builder/AgentBuilderViewModel.swift
+++ b/Convos/Agent Builder/AgentBuilderViewModel.swift
@@ -650,7 +650,7 @@ final class AgentBuilderViewModel: Identifiable {
             withAnimation(.easeInOut(duration: 0.35)) {
                 self.hasCommitted = true
             }
-            // Promote the claimed cache row into a visible conversation
+            // Promote the hidden conversation row into a visible one
             // AFTER `hasCommitted` flips. The `.onChange(of: hasCommitted)`
             // inside `AgentBuilderView` fires the `onCommitted` callback,
             // which the inline-builder host (`ConversationsView`) uses to
@@ -659,11 +659,11 @@ final class AgentBuilderViewModel: Identifiable {
             // `isEmptyCTAActive` flips, the inline builder unmounts, and
             // the onChange handler never fires — leaving no sheet.
             // Sequencing the commit AFTER the state change keeps the
-            // host's callback intact, then the cache flip arrives once
-            // the sheet is already on its way up.
-            if let claimedId = self.newConversationViewModel.claimedConversationId {
-                await self.session.commitClaimedConversation(id: claimedId)
-            }
+            // host's callback intact, then the visibility flip arrives once
+            // the sheet is already on its way up. Covers both the cache-
+            // claimed row and the cache-miss auto-created row (queued until
+            // `.ready` if creation is still in flight).
+            await self.newConversationViewModel.commitConversationVisibility()
         }
 
         Task { @MainActor [weak self] in
--- a/Convos/Conversation Creation/NewConversationViewModel.swift
+++ b/Convos/Conversation Creation/NewConversationViewModel.swift
@@ -739,76 +739,6 @@ class NewConversationViewModel: Identifiable, Hashable {
         }
     }
 
-    /// Promotes this VM's hidden conversation row into a visible one.
-    /// Covers both deferred-visibility shapes: the cache-claimed row
-    /// (id known since acquire) and the auto-created `startsUnused` row
-    /// (id known at `.ready`). When called before the auto-created
-    /// conversation is ready, the commit is queued and flushed by
-    /// `handleStateChange(.ready)` - mirroring how message sends queue
-    /// until the conversation exists.
-    func commitConversationVisibility() async {
-        if let claimedConversationId {
-            // Order behind the `.ready` hook's claim registration so the
-            // commit's claim removal can't be overwritten by a late
-            // registration insert.
-            await claimRegistrationTask?.value
-            await session.commitClaimedConversation(id: claimedConversationId)
-            return
-        }
-        pendingVisibilityCommit = true
-    }
-
-    /// Send a first message through the state machine. Used by wrapping
-    /// flows (e.g. AgentBuilderViewModel) that commit a draft before
-    /// the user sees the chat view. If the state machine hasn't reached
-    /// `.ready` yet, the existing message-stream queue inside
-    /// `ConversationStateMachine.sendMessage` holds the send until it does.
-    func send(text: String) async throws {
-        guard let conversationStateManager else {
-            throw ConversationStateMachineError.noConversationStateManager
-        }
-        try await conversationStateManager.send(text: text)
-    }
-
-    func deleteConversation() {
-        Log.info("Deleting conversation")
-        newConversationTask?.cancel()
-        joinConversationTask?.cancel()
-        // A queued visibility commit must not survive dismissal - the user
-        // abandoned the builder, so a late `.ready` shouldn't promote the
-        // conversation into the chats list.
-        pendingVisibilityCommit = false
-        // Drop the conversation row claimed via `prepareNewConversation()`
-        // when the user backs out without engaging. Key off
-        // `claimedConversationId` so existing-conversation flows
-        // (`.existingConversation(...)`) don't accidentally delete the
-        // user's real conversation. The single-inbox refactor turned the
-        // old per-conversation `session.deleteInbox` cleanup into a no-op
-        // (it would destroy the user's account), so without this the
-        // warm-cached group would persist in the conversations list.
-        // Drafts skip — they don't have a visible row.
-        if let claimedId = claimedConversationId {
-            Task { [session] in
-                await session.discardClaimedConversation(id: claimedId)
-            }
-        }
-    }
-
-    func setDismissAction(_ action: DismissAction) {
-        dismissAction = action
-    }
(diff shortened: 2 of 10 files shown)


yewreeka:
* **#986** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/986?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=986).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

yewreeka:
## Multi-agent review findings and resolutions

Four parallel reviewers (correctness/races, lifecycle/edge cases, conventions/architecture, regression surface) audited the diff. Resolutions:

**Fixed in follow-up commit d0b5739a:**

1. **Unclaimed window for the builder's auto-created stub** (3/4 reviewers, major). Between the pre-publish hidden-row write and the `.ready` claim registration, `consumeUnusedConversationId` could hand the builder's in-flight row to a concurrent compose/Convos-button flow — double-ownership, and the second flow would commit it visible. Fixed structurally: `consumeUnusedConversationId` (and the prewarm dedupe check) now only accept *fully prepared cache rows* — `clientConversationId == id` (builder stubs carry their draft client id and are never recycled through the cache) and a real, non-provisional invite tag. DB-resident, no timing dependence.
2. **`pending-` provisional tag recycling** (lifecycle reviewer, major). An abandoned stub (app killed before the stream echo stored the real tag) could be consumed after restart and its placeholder tag could reach `InviteWriter`. Closed by the same eligibility filter: rows with empty or `pending-` tags never enter the pool.
3. **`pending-` magic string duplicated** (conventions reviewer). Centralized as `provisionalInviteTagPrefix` + `isProvisionalInviteTag(_:)`, mirroring the `DBConversation.draftPrefix` pattern.
4. **Commit-before-register task race** (correctness reviewer, minor). A Make tap landing right after `.ready` could run its commit before the claim-registration task, leaving a stale id in the claim set. The `.ready` hook's work is now held in `claimRegistrationTask` and `commitConversationVisibility()` awaits it first.
5. **Queued commit surviving dismissal** (lifecycle reviewer, minor). Make-before-`.ready` followed by dismiss could still promote the abandoned conversation when `.ready` fired later. `deleteConversation()` now clears `pendingVisibilityCommit`.
6. Doc tweaks: the protocol-extension default for `createConversation(startsUnused:)` is now explicitly marked mock/preview-only; reworded a stale-reading comment in `runPreparation`.

**Accepted with rationale (no code change):**

- **Hidden-row accumulation under sustained publish failure** (regression reviewer, minor): one kept hidden row per failed prewarm cycle, gated by the 30s cooldown, invisible to the user, recyclable for fully-prepared rows. Deleting instead risks the stream echo re-inserting the row visible (libxmtp can throw from `publish()` after the payload landed — observed live during verification). A startup sweep for stale hidden rows is a reasonable follow-up.
- **Make-then-fail-then-retry leaks one hidden stub per failed attempt** (lifecycle reviewer): stubs stay hidden and excluded from the pool; bounded by retry count.
- **`ConversationsCountRepository` counts hidden rows** (regression reviewer, latent): pre-existing behavior (it never filtered `isUnused`), only consumer is a write-only UserDefaults flag. Follow-up candidate.
- **`isAccountOrphaned()` sees the pre-publish row** (regression reviewer, debug-only): direction is non-destructive; only consumer is the orphaned-inbox debug view.

**Verified clean by reviewers:** provisional tags cannot leak into invites or match real tags (`isValidInviteTag` rejects them; both creation paths store the real tag before any `InviteWriter.generate`); all other `createConversation()` callers still create visible rows; chats list / empty-state CTA / pairing gate (`hasAnyUsedConversations`) unaffected; `sendDeviceRemovedSignal` guards against unpublished stubs; cancellation teardown leaves a clean DB; MainActor/Sendable sound.

PR #999, opened 2026-06-05
dev/test: leave the shared Docker stack running and propagate the test exit code

yewreeka:
## Problem

Integration suites (e.g. `LockConversationTests`) flake locally — but the tests themselves are fine: the Lock suite passes 5/5 consecutive runs in ~4.5s against a stable node, even under concurrent load.

The real mechanism, caught live while investigating:

1. The local XMTP compose stack (`convos-ios-*`) is **one shared instance** for every checkout, worktree, and agent session on a machine.
2. `dev/test` ended with an unconditional `compose down`, so whichever session finished first destroyed the node under everyone else's in-flight tests. Those runs then fail wholesale with `Connection refused 127.0.0.1:5556` — or hang forever in suites without a `.timeLimit` trait (orphaned day-old `swift-test` processes were still around).
3. Separately, `dev/test` could never report failure: `swift test --skip-build || true` followed by `test_exit_code=$?` captures the exit code of `|| true`, so the script always exited 0 even when tests failed.

## Changes

- **Leave the Docker stack running by default**; `--down` opts into teardown. The build-failure path also no longer tears the stack down.
- **Skip `dev/up` when the node is already serving on 5556** — `compose pull` + `up -d` can recreate the node container when a newer image exists, which is the same class of mid-run disruption.
- **Fix the exit-code capture** so test failures propagate to the caller.
- **Update CLAUDE.md** testing guidance to stop instructing sessions to run `./dev/down` as routine cleanup, and document the shared-stack hazard.

## Test plan

- `bash -n` clean; verified the `set -e`-safe exit-code construct (`false || ec=$?` → 1, `true || ec=$?` → 0)
- `./dev/test --help` and unknown-flag paths exit 0/1 as expected
- Full `./dev/test` run end-to-end with the new script: reused the already-running node, **1104 tests in 158 suites passed**, stack left running afterwards

> [!NOTE]
> ### Leave the shared Docker stack running by default and propagate test exit codes in `dev/test`
> - The [dev/test](https://github.com/xmtplabs/convos-ios/pull/999/files#diff-20c04bb8529d71d3c6a1a55e5cdf47f9737a908e2592f3f85594fbd71cfb45a0) script now checks if an XMTP node is already serving on `localhost:5556` (via `nc -z`) before starting containers, reusing it if found.
> - The Docker stack is no longer torn down after a test run by default; pass `--down` to stop it explicitly.
> - Build failures no longer trigger automatic teardown.
> - Test exit codes are now captured via `swift test --skip-build || test_exit_code=$?` instead of being swallowed by a pipe to `true`.
> - [CLAUDE.md](https://github.com/xmtplabs/convos-ios/pull/999/files#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7) is updated to reflect the shared stack behavior and remove `./dev/down` from routine cleanup examples.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ec86678.</sup>
> 
>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -549,19 +549,18 @@ cp -r ~/Downloads/SamplePhotoPickerApp claude-code-resources/
 Use the `./dev/test` script for running tests. **Most tests require Docker** for the local XMTP node:
 
 ```bash
-# Full test suite (starts Docker automatically)
+# Full test suite (starts Docker automatically, leaves the stack running)
 ./dev/test
 
-# Start Docker manually, run tests, then stop
+# Start Docker manually if needed, then run tests
 ./dev/up
 swift test --package-path ConvosCore
-./dev/down
 
 # Run a single test
-./dev/up  # Start Docker first
 swift test --filter "TestClassName" --package-path ConvosCore
-./dev/down  # Stop when done
 ```
+
+**The Docker stack is shared.** Every checkout, worktree, and agent session on the machine uses the same compose project, so do not run `./dev/down` (or `./dev/test --down`) as routine cleanup — tearing the stack down kills any other session's in-flight integration tests, which then fail with `Connection refused 127.0.0.1:5556` or hang. Leave the stack running; only stop it when you know nothing else is using it.
 
 **Docker is required.** If Docker is not running, start it with `./dev/up` before running tests. If Docker fails to start (not installed, daemon not running, port conflicts), **do not skip the tests** — stop and notify the user that Docker cannot start and tests cannot run. Never push untested code.
 
--- a/dev/test
+++ b/dev/test
@@ -1,20 +1,32 @@
 #!/bin/bash
 set -eou pipefail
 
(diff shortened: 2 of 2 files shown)


yewreeka:
* **#999** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/999?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/999?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=999).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #949, opened 2026-06-03
Enable production observability: on-device logging + Sentry crash reporting

yewreeka:
## What

Production builds get observability through two complementary channels (both previously disabled):

**On-device logging**
- `ConvosLog.configure` no longer early-returns for `.production` - the file + console log handlers are installed in every environment. The file handler records `.info` and above (its default level; nothing in the codebase raises it to `.debug`), capped at 10MB with rotation.
- The libxmtp persistent log writer is now active in every environment, with an environment-gated level: `.debug` for dev/local (unchanged), **`.warn` for production**. Hourly rotation, 10 files max.
- The conversation info screen now shows the **share-logs row in production** (under a "Diagnostics" header) so users can send on-device logs to support via a share sheet. The internal-only debug rows (fork status, commit logs, restore invite tag) remain hidden in production.

**Sentry crash + error reporting in production**
- `SentryConfiguration` now enables Sentry for `.production` with privacy-safe options: no screenshots, no view hierarchy, no default PII - crash/error reports with stack traces only. Dev builds keep their richer debugging context (unchanged).
- `generate-secrets-secure.sh` accepts the `Prod`/`Release` build configurations and bakes in `SENTRY_DSN_PROD` (empty when unset, which leaves Sentry disabled - safe default).
- `testflight-prod.yml` wires `SENTRY_DSN_PROD` and uploads dSYMs to Sentry after the build so prod crash reports are symbolicated (skips cleanly when secrets are unset). `sentry-cli` added to the nix dev shell.
- `firebase-pr.yml` now passes `SENTRY_DSN_DEV` - dev CI builds were silently shipping an empty DSN, so Sentry was effectively off in dev CI builds too.

## Why

Production builds emitted no diagnostics - app logs off, libxmtp logs off, Sentry dev-only. That made prod-only failures impossible to debug (see the "New Convo fails every time" report this branch was used to diagnose).

This now provides two tiers: Sentry reports crashes and errors remotely from every prod device (symbolicated), and on-device logs provide deeper context when a device is available, at a level that keeps the volume and privacy footprint small.

## How the code-review concerns were addressed

| Concern | Resolution |
|---|---|
| Debug-level PII written on every prod device | App file log records `.info`+ (not `.debug`); libxmtp gated to `.warn` in prod. Logs leave the device only through the user-initiated share-logs row in conversation info; nothing is transmitted automatically, and the prod Sentry config sends no screenshots/view-hierarchy/PII. |
[...]

PR #1001, opened 2026-06-05
fix(media): bound memory in the receive-side image pipeline (background OOM crash)

yewreeka:
## Why

A TestFlight crash (2.0.0 build 914, `EXC_BAD_ACCESS` in `_platform_memset`) turned out to be an out-of-memory failure in disguise: the app was running **in the background** (`Role: Non UI`) when a ~900-byte allocation inside CFNetwork's `HTTPProtocol::_instantiateProtocol` returned NULL and Apple's code bzero'd `NULL + 0x10` without checking. Disassembly of CFNetwork confirms the pattern, and the crash registers (`x0=0x10`, `x2=0x388`) match exactly.

A process that can't allocate 900 bytes has truly exhausted its memory budget — and background jetsam budgets are far smaller than foreground. The crash-time threads point at the media pipeline: one thread mid-`Data.write(to:options:.atomic)` (image cache disk write), another starting the next HTTP request.

## What was wrong

The receive-side media pipeline had three compounding unbounded-memory behaviors:

1. **Full in-memory buffering of downloads.** `EncryptedImageLoader` and `ImageCache.fetchImageFromNetwork` used `URLSession.shared.data(from:)`, holding the entire encoded payload in RAM — for encrypted images, ciphertext *and* plaintext simultaneously.
2. **Unbounded decode of sender-controlled payloads.** Five call sites decoded network/disk bytes with `UIImage(data:)`, which decodes at full resolution. A 20k x 20k JPEG decodes to ~1.6 GB; even honest 12 MP photos decode to ~48 MB each. Senders outside our upload pipeline (which caps at 2048px) can send anything.
3. **Unbounded fan-out during sync.** `ConversationWriter` fires a detached `Task` per conversation for group-image prefetch; the avatar prefetcher caps per-call but not process-wide. A background sync over many conversations stacks N concurrent downloads, each holding full buffers.

## The fix

- **`BoundedImageDecode`** (new, cross-platform): decodes via ImageIO's thumbnail API with a hard cap (default 2048px, matching the sender-side attachment cap) on the longest edge, reading from data or straight from a file. Replaces `UIImage(data:)` at all five receive-side decode sites in `ImageCache`, including disk loads.
- **`EncryptedImageLoader`** now streams ciphertext to a temp file (`URLSession.download(from:)`), rejects payloads over 20 MB, decrypts from a memory-mapped read, and bounds process-wide concurrency with a new **`AsyncSemaphore`** (width 4). Every encrypted-image call site (writers, prefetcher, inline cache fetch) funnels through it, so sync bursts are globally capped.
- **`ImageCache.fetchImageFromNetwork`** streams to a temp file instead of buffering, then bounded-decodes from the file.
- **`ImageCache.loadImageFromDisk`** decodes directly from the file URL instead of reading encoded bytes into memory first.

Worst-case per-image footprint drops from *(encoded payload + unbounded decoded bitmap)* to *(bounded ≤2048px bitmap ≈ 17 MB max)*, and at most 4 encrypted downloads hold buffers at once regardless of how many conversations sync.

## Tests

- `BoundedImageDecodeTests` (cross-platform, runs on macOS CI): oversized clamp, aspect preservation, small-image passthrough, custom cap, file-URL variant, invalid data/missing file.
- `AsyncSemaphoreTests`: concurrency bound holds under 20 contending tasks, slot released on throw, value passthrough.
- Full ConvosCore suite green locally (Docker XMTP node); app target builds clean for iOS sim under the 100 ms type-check budget.

## Out of scope

- `RemoteAttachment.content()` (XMTP SDK) still buffers attachment downloads in memory — needs an upstream SDK change.
- Sentry `enableWatchdogTerminationTracking` for OOM visibility is covered by the Sentry rollout (#949).

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1001"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783272298&installation_id=128169237&pr_number=1001&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1001&signature=70144e1630789f19fc97a978f301635b50bde61fc685660e6bd0d4560432ffd4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Bound memory in the receive-side image pipeline to fix background OOM crashes
> - Adds [`BoundedImageDecode`](https://github.com/xmtplabs/convos-ios/pull/1001/files#diff-06876b1e7f29ec8a732eee168fe20f8f5cd5db16aa65f8f578f1f99ed5db6b1d) which uses ImageIO thumbnail APIs to clamp decoded bitmaps to 2048px on the longest edge; replaces all `UIImage(data:)` calls in `ImageCache` and `EncryptedImageLoader`.
> - Adds [`AsyncSemaphore`](https://github.com/xmtplabs/convos-ios/pull/1001/files#diff-90285d486b3e9ceb05045be358012d5575fc0da34cecbf92ed9984c5e23d1b78), a FIFO async counting semaphore, used to cap concurrent encrypted-image downloads to 4 per priority tier (interactive vs. background).
> - Refactors [`EncryptedImageLoader`](https://github.com/xmtplabs/convos-ios/pull/1001/files#diff-7b33fc0530e99a3e9a50f447631666ada9ade550b718395a93709803597b28bd) to stream ciphertext to a temp file via `URLSession.download`, reject payloads >20 MB, and decrypt from a memory-mapped file instead of an in-memory buffer.
> - Refactors [`ImageCache`](https://github.com/xmtplabs/convos-ios/pull/1001/files#diff-445c3a2d9023d5d29469046555d79fa387c40b98bfcf772c84cdf1679fc38b91) network fetches to use `URLSession.download` and disk cache reads to decode directly from file URL, avoiding full in-memory buffering of encoded payloads.
> - Behavioral Change: decoded image dimensions are now capped at 2048px; images larger than 20 MB ciphertext are rejected with `URLError(.dataLengthExceedsMaximum)`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized aad24cb.</sup>
> 
>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift
+++ b/ConvosCore/Sources/ConvosCore/Image Cache/ImageCache.swift
@@ -527,8 +527,9 @@ public final class ImageCache: ImageCacheProtocol, @unchecked Sendable {
                 Log.error("Invalid URL for cacheAfterUpload: \(url), caching without URL tracking")
             }
 
-            // Create image from data for memory cache
-            guard let image = UIImage(data: imageData) else {
+            // Create image from data for memory cache (bounded decode:
+            // the data came off the network and is sender-controlled)
+            guard let image = BoundedImageDecode.image(from: imageData) else {
                 Log.error("Failed to create UIImage from data: \(identifier)")
                 return
             }
@@ -567,7 +568,10 @@ public final class ImageCache: ImageCacheProtocol, @unchecked Sendable {
             }
 
             do {
-                let (data, response) = try await URLSession.shared.data(from: url)
+                // Stream to a temporary file so the encoded payload is never
+                // fully buffered in memory, then decode with a bounded size.
+                let (tempFileURL, response) = try await URLSession.shared.download(from: url)
+                defer { try? FileManager.default.removeItem(at: tempFileURL) }
 
                 // Validate response
                 guard let httpResponse = response as? HTTPURLResponse,
@@ -576,7 +580,7 @@ public final class ImageCache: ImageCacheProtocol, @unchecked Sendable {
                     return nil
                 }
 
-                guard let image = UIImage(data: data) else {
+                guard let image = BoundedImageDecode.image(contentsOf: tempFileURL) else {
                     Log.error("Failed to decode image from URL: \(url)")
                     return nil
                 }
@@ -631,7 +635,7 @@ public final class ImageCache: ImageCacheProtocol, @unchecked Sendable {
                 groupKey: key
             )
 
-            guard let image = UIImage(data: decryptedData) else {
+            guard let image = BoundedImageDecode.image(from: decryptedData) else {
                 Log.error("Failed to create UIImage from decrypted data: \(identifier)")
                 return nil
             }
@@ -749,7 +753,7 @@ public final class ImageCache: ImageCacheProtocol, @unchecked Sendable {
     // MARK: - Persistent Storage (chat photo attachments)
 
     public func cacheData(_ data: Data, for identifier: String, storageTier: ImageStorageTier) {
-        guard let image = UIImage(data: data) else {
+        guard let image = BoundedImageDecode.image(from: data) else {
             Log.error("Failed to create UIImage from data for persistent cache: \(identifier)")
             return
         }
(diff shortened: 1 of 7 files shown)


yewreeka:
## Review-matrix follow-up (commit aad24cba)

A 5-dimension review matrix (correctness / conventions / performance / tests / security, each finding adversarially verified) confirmed 5 issues; all are addressed in this commit:

1. **Priority inversion (perf, major)** — encrypted-image downloads now use separate `interactive` / `background` gates (4 slots each). The UI-blocking cold-cache path (`loadImage → fetchEncryptedImageInline`) takes `.interactive` and no longer queues FIFO behind sync's prefetch burst. Pinned by a test that saturates the background gate and asserts an interactive fetch still completes.
2. **Serialized disk decodes (perf, major)** — `loadImageFromDisk` now only checks existence + touches access dates on the serial disk queue; the bounded decode runs on the caller's task, so concurrent avatar loads decode in parallel.
3. **Untestable loader (tests, critical)** — extracted `decryptDownloadedFile(at:response:params:)` as an internal seam + injectable download transport. New tests cover the decrypt round-trip, 20MB cap, non-2xx, and non-HTTP responses.
4. **Gate wiring untested (tests, major)** — new test fires 12 concurrent loads through a stub transport and asserts peak in-flight ≤ 4; removing `withSlot` now fails the suite.
5. **EXIF transform untested (tests, major)** — new oriented-JPEG fixtures verify rotation is baked into pixels and clamping applies to post-transform dimensions.

Verified non-issues (claims refuted by adversarial verifiers): full-res photo loss, mmap-after-unlink integrity, ciphertext-on-temp-disk regression, missing data-protection class, plaintext-path DoS asymmetry.

Full suite: 1121 tests / 162 suites green; iOS app target builds clean.

PR #1010, opened 2026-06-06
Fix conversation-open flicker: keep bottom anchor pinned while the open transition settles

yewreeka:
## What

Fixes the conversation-open flicker that survived #987/#998, plus the warm-reopen "ride" where the composer and the bottom-pinned message list visibly slide up for the length of the push spring. Also fixes the agent contact-card cell's height estimation.

## Root causes (all confined to the open-transition window)

1. **Bar inset deferred mid-transition.** The composer's height measurement arrives inside a `performWithoutAnimation` scope, so `scheduleBottomBarInsetUpdate`'s `areAnimationsEnabled` check (#987) routed it to the deferred path — the list painted anchored against a stale inset, then snapped (+227pt in one frame, measured) when the initial load's completion flushed. The settling check now precedes the deferral and applies via `applyBottomInsetInstantly` (#998's path, plain property sets).
2. **Safe-area mount never re-anchored.** The SwiftUI bottom bar registers in the safe area a render pass after the list's first bottom anchor. New `viewSafeAreaInsetsDidChange` override re-anchors arithmetically while settling.
3. **Inherited animation on "non-animated" anchors.** `scrollToBottom(animated: false)` ended in a naked `contentOffset` assignment; during open it runs inside the push transition's animation scope, so CA animated the re-anchor with the transition spring — the whole bottom assembly rode the bar's entrance ~110pt for ~500ms on every warm re-open (cold opens mask it behind main-thread load). Now wrapped in `performWithoutAnimation` (same for `flushPendingComposerInset`).
4. **Transient safe-area double-count.** For a few frames the bar contributes to the safe area *and* to `contentInset` (97+83=180 vs settled 131); anchoring against it over-pinned the list, which then stepped down 49pt at push end. New `bottomAnchorInset` caps the settle-time anchor at the settled target (bar inset + window safe area).
5. **Out-of-band growth skip during initial self-sizing.** #982's >12pt growth skip fired for initial estimated-to-real cell sizing (one cell grew 601pt past its estimate) while its reveal callback is gated off during `.loadingInitialMessages`/`.updatingCollection` — nobody re-pinned. The layout now compensates all self-sizing growth until `viewDidAppear`, and the reveal path is scoped to last-item growth (the send/new-media case it was built for), so growth above the viewport no longer shifts content under a reader.
6. **Height estimates.** `estimatedHeight` ignored `agentContactCard` entirely (16pt baseline vs ~180pt real) and used a flat 40pt for `.text` regardless of length (long replies render at 400–500pt). Card height is now estimated and text scales by character count, shrinking every self-sizing delta the anchor has to absorb.

## Validation

- Instrumented builds with offset/inset logging at every anchor point: every paint now pins to the exact bottom for the current inset state; the initial-load completion's scroll is a no-op; zero post-`viewDidAppear` offset jumps across repeated open/close cycles.
- Frame-by-frame video measurement (composer-edge tracking) on a photo-heavy conversation: before, 45+ frames of upward creep per warm open; after, content and composer land at final position within ~5 frames and never move — 3/3 consecutive warm opens.
- Send-path smoke test: multi-line composer collapse + optimistic append still settle in one motion (steady-state #982 behavior unchanged after `viewDidAppear`).

> [!NOTE]
> ### Fix conversation-open flicker by keeping the bottom anchor pinned during the open transition
> - During initial layout settling, bottom bar inset updates are applied synchronously and safe area inset changes trigger an immediate re-anchor to bottom, preventing the viewport from jumping.
> - Adds `compensatesAllSelfSizingGrowth` to `MessagesCollectionLayout` (default `true`) so all self-sizing growth is immediately compensated during settling; after `viewDidAppear`, this reverts to `false` to resume normal out-of-band reveal behavior.
> - Adds `bottomAnchorInset` to cap the anchor inset during settling, avoiding double-counting transient safe area contributions from the transition.
> - Non-animated `scrollToBottom` calls are wrapped in `UIView.performWithoutAnimation` to prevent inheriting ambient transition animations.
> - Improves estimated cell heights in `DefaultMessagesLayoutDelegate` — text messages now scale with character count and groups with an `agentContactCard` report a larger initial estimate — reducing first-layout underestimation that contributed to the flicker.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3a03a3d.</sup>
> 
>

### DIFF
--- a/Convos/Conversation Detail/Messages/Messages Collection Layout/MessagesCollectionLayout.swift
+++ b/Convos/Conversation Detail/Messages/Messages Collection Layout/MessagesCollectionLayout.swift
@@ -145,6 +145,15 @@ open class MessagesCollectionLayout: UICollectionViewLayout {
 
     private var hasPendingOutOfBandGrowthNotification: Bool = false
 
+    /// While the owning controller's open transition is settling, initial
+    /// self-sizing of estimated cells produces large out-of-band growth whose
+    /// reveal callback is gated off (the controller is still loading or
+    /// applying the initial update). Skipping compensation then leaves the
+    /// bottom of the list below the fold until the load completion snaps it -
+    /// the conversation-open flicker. The controller sets this true until
+    /// `viewDidAppear`, restoring the immediate compensation for that window.
+    var compensatesAllSelfSizingGrowth: Bool = true
+
     private func notifyOutOfBandBottomGrowth() {
         guard !hasPendingOutOfBandGrowthNotification else { return }
         hasPendingOutOfBandGrowthNotification = true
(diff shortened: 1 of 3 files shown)


yewreeka:
* **#1010** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/1010?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/1010?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=1010).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #1013, opened 2026-06-07
Render avatars without GeometryReader when the size is known

yewreeka:
## What

Fav'd/pinned conversations scroll worse than the rest of the Chats list. The pinned section renders large (96pt) avatars, and group pins use the clustered avatar (one sub-avatar per member). Every avatar view (`EmojiAvatarView`, `MonogramView`, `ClusteredAvatarView`) wrapped its content in a `GeometryReader` to read its own side length — so a single clustered pinned cell stacked up to ~8 nested `GeometryReader`s, each re-running its content closure on every layout/measure pass during scroll.

## Change

The avatar size is already known at these call sites (the pinned cell fixes it at 96pt; the cluster computes each sub-avatar's size). Thread that size down so the avatar views render at an exact side length **without** a `GeometryReader` — mirroring `MessageAvatarView`, which already avoids `GeometryReader` for exactly this scroll-perf reason.

- `EmojiAvatarView`, `MonogramView`, `ClusteredAvatarView`: render geometry-free when a `size` is provided; keep the `GeometryReader` self-sizing path when it's nil.
- `AvatarView` / `ProfileAvatarView` / `ConversationAvatarView`: forward the size to those fallbacks.
- `PinnedConversationItem`: pass its fixed avatar size.

Other surfaces (list rows that only constrain with an outer `.frame`) are untouched — they keep the self-sizing path. Visuals are identical: same font/padding ratios, just fed an explicit side instead of a measured one (verified on the simulator).

## Diagnosis

Time Profiler on a seeded clustered-pin list (5 fav'd groups with multi-member image avatars, pinned via direct DB writes since invite-join is currently flaky on dev) showed `sizeThatFits` / `Layout` dominating the pinned section's render cost — consistent with the nested `GeometryReader`s.

## Validation note (please read)

This is a **structural** reduction (removes up to ~8 nested `GeometryReader`s per clustered pinned cell), verified to preserve appearance. It is **not yet confirmed against the on-device hitch**:
- The simulator (M-series Mac) renders too fast to reproduce the dropped frames — three hypotheses (snapshot-feedback loop, raw clustered-cell cost, GeometryReader) all showed no measurable per-frame delta in-sim, and a snapshot-rebuild-per-frame theory was instrumented and disproven.
- On-device Animation Hitches capture was blocked this session: one iPhone's instruments server returned "an unknown problem is preventing this device from recording"; the other device lacked the pinned test data.

Landing this as a sound, low-risk improvement that directly targets the measured layout cost. An on-device before/after Animation Hitches comparison is the right follow-up to confirm the frame-drop reduction.

> [!NOTE]
> ### Render avatars without GeometryReader when explicit size is known
> - Adds an optional `size` parameter to `MonogramView`, `EmojiAvatarView`, `ClusteredAvatarView`, `AvatarView`, `ProfileAvatarView`, and `ConversationAvatarView`.
> - When `size` is provided, each view renders directly at the given dimensions using an extracted `circle(side:)` or `clusterContent(side:)` helper, skipping the `GeometryReader` path entirely.
> - Call sites in `PinnedConversationItem` pass `avatarSize` explicitly to `ConversationAvatarView` to take advantage of this path.
> - Behavioral Change: views with a known size no longer go through `GeometryReader`, which changes the layout measurement lifecycle for those call sites.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f3a65f9.</sup>
> 
>

### DIFF
--- a/Convos/Conversations List/PinnedConversationItem.swift
+++ b/Convos/Conversations List/PinnedConversationItem.swift
@@ -29,7 +29,7 @@ struct PinnedConversationItem: View {
 
     var body: some View {
         VStack(alignment: .center, spacing: DesignConstants.Spacing.step2x) {
-            ConversationAvatarView(conversation: conversation, conversationImage: nil)
+            ConversationAvatarView(conversation: conversation, conversationImage: nil, size: avatarSize)
                 .frame(width: avatarSize, height: avatarSize)
                 .padding(.top, DesignConstants.Spacing.stepX)
                 .overlay(alignment: .top) {
@@ -144,7 +144,7 @@ struct PinnedConversationContextPreview: View {
                     )
             }
 
-            ConversationAvatarView(conversation: conversation, conversationImage: nil)
+            ConversationAvatarView(conversation: conversation, conversationImage: nil, size: avatarSize)
                 .frame(width: avatarSize, height: avatarSize)
 
             HStack(spacing: DesignConstants.Spacing.stepX) {
--- a/Convos/Shared Views/AvatarView.swift
+++ b/Convos/Shared Views/AvatarView.swift
@@ -8,6 +8,10 @@ struct AvatarView: View {
     let placeholderEmoji: String?
     let placeholderImageName: String?
     let agentVerification: AgentVerification
+    /// Forwarded to the emoji/monogram fallbacks so they render without a
+    /// `GeometryReader` when the caller already knows the avatar size (see
+    /// `EmojiAvatarView.size`). Nil keeps the self-sizing path.
+    let explicitSize: CGFloat?
     @State private var cachedImage: UIImage?
 
     init(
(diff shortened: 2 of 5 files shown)


yewreeka:
* **#1013** <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/1013?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/xmtplabs/convos-ios/1013?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `dev`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/xmtplabs?ref=merge-queue-instructions-in-stack-comment&prId=1013).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #1037, opened 2026-06-11
Fix local-stack onboarding friction found during a fresh-machine /run local

yewreeka:
## Summary

Ran `/run local` end-to-end on a fresh machine (all xmtplabs repos newly cloned) and fixed every onboarding friction point hit along the way:

- **`stack.sh` init bug** — backticks in the "wrote stack.env" message were executed as command substitution, so `make init` accidentally ran `make bootstrap` mid-init, surfacing as a confusing "Docker not running / bootstrap Error 1" on every fresh machine.
- **Wrong 1Password vault in `stack.env.example`** — the shipped `OP_*` refs pointed at a nonexistent `Engineering` vault. The team items live in the **Convos** vault, and Dev/Local share the single "Firebase App Check Debug Token" item (same ref `/firebase-token` uses).
- **Silent, permanent secret failure** — bootstrap before `op signin` left a skeleton `.dev.vars` (no `CONVOS_API_KEY`) and never retried 1Password on later runs. Now it warns up front when the CLI isn't signed in and re-fetches when the file is missing or incomplete.
- **Herald Node-ABI crash** — herald pins node `>=25` and ships `better-sqlite3`; installing under the caller's (older) node builds the wrong ABI and herald dies at startup with `ERR_DLOPEN_FAILED`. Deps now install via a login shell, the same env `start_svc` launches services with.
- **Docker not running just died** — `bootstrap`/`up`/`hermes-build` now auto-start OrbStack or Docker Desktop on macOS and wait; `doctor` warns instead of aborting its remaining checks, and only nags about the CPU cap on Docker Desktop (OrbStack allocates dynamically).
- **`make init` cloned duplicate repos** — when all four service repos already sit next to convos-ios, init now defaults the workspace to that parent dir instead of cloning copies into `<parent>/convos-stack`.
- **Build failures masked as success** — the `xcodebuild ... | tail` snippets in `run.md` return exit 0 on `** BUILD FAILED **`. Added `set -o pipefail` and a note.

## Test plan

- [x] Full `/run local` flow on a fresh machine: `make init` (reused existing clones) → `bootstrap` → `up` → build/install/launch "Convos (Local)" on a dedicated simulator
- [x] All services healthy: backend/herald/worker/MinIO 200, Postgres healthy, Hermes ready, credits seeded
- [x] Verified under both Docker Desktop and OrbStack (engine auto-start + doctor behavior)
- [x] Herald recovers from the ABI mismatch with the login-shell install
- [x] `bash -n` on `stack.sh`; no Swift code touched (docs + shell only)

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1037"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783793534&installation_id=128169237&pr_number=1037&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1037&signature=b85b5520bcf4b5348995fe436ea53a774e1369727bbb5bf931febc44250b4aeb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Fix local-stack onboarding friction for fresh-machine setup
> - Adds `ensure_docker` to auto-start OrbStack or Docker Desktop on macOS and wait up to ~120s before failing, replacing immediate failures when Docker is not running
> - Adds `op_ready` helper and warns (rather than silently proceeding) when 1Password CLI is not signed in during `bootstrap`
> - Adds `guard_port` to abort early if a service port is already held by an unrelated process, preventing false-healthy states
> - Fixes `svc_running` to verify the recorded PID's working directory matches the expected service directory, reducing false positives from PID reuse
> - Fixes `bootstrap_herald` to install dependencies via a login shell first, reducing native module ABI mismatches
> - Fixes `bootstrap_assistants` to refetch `.dev.vars` when `CONVOS_API_KEY` is missing and use a temp file to avoid leaving a broken config after a pre-signin bootstrap
> - Updates `stack.env.example` vault references from `Engineering` to `Convos` and clarifies documentation in `run.md` and `local-stack/README.md`
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7bda62c.</sup>
> 
>

### DIFF
--- a/.claude/commands/run.md
+++ b/.claude/commands/run.md
@@ -31,7 +31,7 @@ Bundle IDs by scheme:
 Skip this whole step for Dev/Prod schemes. The orchestration is committed in this repo at `dev/local-stack/` and resolves its external workspace itself (via the gitignored `.convos-stack` pointer or `$CONVOS_REPOS_DIR`).
 
 1. `LS="$(git rev-parse --show-toplevel)/dev/local-stack"`.
-2. **First-run:** `make -C "$LS" status` (30s). If it errors *"no workspace configured — run: make init"*, the machine isn't set up: run `make -C "$LS" init` (defaults the workspace to a sibling of this repo and clones the service repos; ~600000ms timeout), then **pause** and tell the user to set the `OP_*` 1Password refs in `<workspace>/stack.env` and run `make -C "$LS" bootstrap` once.
+2. **First-run:** `make -C "$LS" status` (30s). If it errors *"no workspace configured — run: make init"*, the machine isn't set up: run `make -C "$LS" init` (workspace defaults to a sibling of this repo — or to the parent dir itself when the service repos are already cloned there — and clones any missing service repos; ~600000ms timeout). The written `stack.env` ships working `op://Convos/...` refs, so next run `make -C "$LS" bootstrap` once. If bootstrap warns the 1Password CLI is not signed in, **pause** and ask the user to run `op signin` (account xmtpinc) or enable 1Password Settings > Developer > "Integrate with 1Password CLI", then re-run bootstrap (it re-fetches `.dev.vars` if it's missing or incomplete).
 3. **Bring it up if needed:** if `status` shows backend/worker down, run `make -C "$LS" up` with a **1200000ms (20 min)** timeout (first run builds the Hermes image — capped; later runs are fast). Relay any Docker-cap warning from `make -C "$LS" doctor`.
 4. **Configure this checkout as a thin client:** `make -C "$LS" ios-config IOS="$(pwd)"` — sets `config.local.json` `xmtpNetwork→dev` and writes `./.env` (shared Firebase Local token; `CONVOS_API_BASE_URL` left empty so Local auto-detects the Mac LAN IP and a Dev build on the same checkout isn't redirected to localhost).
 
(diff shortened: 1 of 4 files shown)


yewreeka:
Pushed a second commit after debugging the live stack — three more failure modes the orchestration couldn't see:

- **Port squatters**: a backend from an old checkout (running since June 1) held :4000 and answered all health checks, so the stack looked green while the app got 401 \"Invalid AppCheck token\" from a stale database. `cmd_up` now dies with the squatter's pid/command instead of silently being shadowed.
- **Dead service children behind live pidfiles**: `svc_running` was a bare `kill -0`; pid reuse or a wrapper whose server child died with EADDRINUSE made `make up` skip the start and spin in a 60s health-check loop. It now verifies the pid's cwd matches the service dir.
- **False \"ok App Check disabled\"**: `disable_app_check` used curl without `-f`, so an HTTP 401 still reported success.

Verified live: idempotent re-up recognizes all running services, App Check disable round-trips, app authenticates and syncs against the local stack.

PR #1041, opened 2026-06-11
Hold builder-bundle publish until the agent has joined the conversation

yewreeka:
## Summary

The agent-builder brief (manifest + attachments + prompt text) was published to XMTP **before the agent joined the group**, so the agent could never read it (XMTP members can't see messages published before their join):

- The existing-conversation flow sent the bundle first and called `requestAgentJoin` after.
- The home flow requests the join at `.ready`, but the join is async, so the publish could still race the actual membership.

## Fix (two layers)

1. **`AgentBuilderViewModel.commitToExistingConversation`** — request the agent join *before* persist/uploads/send (matches the home flow's ordering, and gives the backend a head start while uploads finish).
2. **`OutgoingMessageWriter.sendBuilderBundleImpl`** — gate the publish on a **verified agent member** appearing in the conversation (GRDB `ValueObservation`, instant if already joined), with a 150s timeout fallback matching the join polling window so a failed/slow join still publishes rather than stranding the brief. The gate sits between local prepare (optimistic rows, hidden from the sender's UI) and network publish, so nothing visibly stalls and the manifest still references the real XMTP ids.

This closes the race in both flows regardless of how fast the agent joins.

## Test plan

- [x] Full ConvosCore suite: 1148 tests in 166 suites, all passing
- [x] Full app build (Convos (Local)) compiles
- [x] SwiftLint clean on both files
- [ ] Manual: build an agent via the builder against the local stack and confirm (via `convos.log`) the publish happens after the member-added event

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1041"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783799559&installation_id=128169237&pr_number=1041&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1041&signature=60d5e20742cdbab01e2b741854b57e07182eb428b79eedb0996b12fb50eeb84c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Hold builder bundle publish until an agent has joined the conversation
> - Before publishing a builder bundle in an existing conversation, `OutgoingMessageWriter.sendBuilderBundleImpl` now waits for an agent to appear as a current conversation member, preventing the agent from missing content it cannot read after joining.
> - A new `waitForAgentMember` helper observes the database for agent membership and falls back to polling, with a 150-second timeout before proceeding anyway.
> - In `AgentBuilderViewModel`, the agent join is requested before persisting/sending (previously after), and the join task is detached from the builder VM so sheet dismissal no longer cancels it.
> - A new `awaitsAgentJoin` parameter (default `true`) threads through `sendBuilderBundle` at every layer from `OutgoingMessageWriterProtocol` down to the state machine and mocks.
> - Risk: if the agent never joins, publish is delayed up to 150 seconds before proceeding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0d4a37d.</sup>
> 
>

### DIFF
--- a/Convos/Agent Builder/AgentBuilderViewModel.swift
+++ b/Convos/Agent Builder/AgentBuilderViewModel.swift
@@ -746,9 +746,29 @@ final class AgentBuilderViewModel: Identifiable {
         // owns its own message writer, so it sends independently of the
         // dismissed builder.
         Task { @MainActor [innerVM, voiceMemoSnapshot, summaryToPersist, conversationIdForPersist, textMessageId, bundleMessageId, session] in
-            // Persist the summary first so the chat the user returns to renders
-            // the card immediately (its `summaryPublisher` picks this up) and
-            // the filter set is on disk before the bundle lands.
+            // Request the agent join before anything else: the bundle publish
+            // below holds until the agent is a verified member (XMTP members
+            // can't read messages published before they joined), so the join
+            // must already be in flight by the time the send runs -- and the
+            // head start lets the backend add the agent while the summary
+            // persists and uploads finish. The join must finish even after
+            // the builder sheet dismisses. Unlike the draft flow, there is no
+            // draft to discard -- the user's group stays -- so the join
+            // survives the view closing (like `ConversationViewModel`'s
+            // committed-conversation join, the opposite of the draft path).
+            let slug = innerVM.conversation.invite?.urlSlug ?? ""
+            if slug.isEmpty {
+                Log.warning("AgentBuilder(existing): no invite slug; skipping agent join")
+            } else {
+                do {
+                    _ = try await session.requestAgentJoin(slug: slug, options: .agentBuilder)
+                } catch {
+                    Log.error("AgentBuilder(existing): requestAgentJoin failed: \(error.localizedDescription)")
+                }
+            }
+            // Persist the summary before the send so the chat the user returns
+            // to renders the card immediately (its `summaryPublisher` picks
+            // this up) and the filter set is on disk before the bundle lands.
             do {
                 try await session.agentBuilderSummaryWriter().save(summaryToPersist, for: conversationIdForPersist)
             } catch {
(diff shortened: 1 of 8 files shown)


yewreeka:
Ran a multi-agent adversarial review matrix over the first cut (5 dimensions × 3-skeptic verification, 83 agents): 18 confirmed findings, including one **critical** — the original gate guarded nothing. XMTP `prepare` queues a libxmtp send intent immediately, and any concurrent publish or group sync flushes all pending intents (a read receipt fires when the user lands in the chat after Make; the agent-join polling itself syncs every 5s and publishes intents before receiving). The held brief was still reaching the network pre-join, encrypted in the pre-join epoch, with the later explicit publish silently no-oping.

Second commit addresses the confirmed findings:

- **Gate moved before any `prepare`** — no part of the brief exists in the XMTP store until the agent can read it
- **Predicate = current agent member** (`DBConversationMember` ⋈ `isAgent` profile) — stale profile rows for removed agents no longer open the gate; attestation verification no longer required (membership is what gates readability)
- **`awaitsAgentJoin` parameter** — existing-conversation flow passes `false` when the join was never requested or failed, so the bundle publishes immediately instead of stalling 150s
- **Home flow detaches `agentJoinTask` on commit** — sheet teardown can't cancel the join mid-flight anymore
- **Observation failure → polling fallback** with honest logging
- **7 new unit tests** pin the gate semantics (opens on existing member, opens mid-wait, ignores stale profiles/other conversations, times out and falls through)

Known residuals (documented in code): a brief for a *second* agent in a conversation that already has one can still race — `AgentJoinResponse` carries no agent identity to wait on (needs backend support); app-kill during the hold loses the brief (no stranded "sending" rows though, since nothing is prepared until the gate opens).

Verified: 1155 tests green, full app build, SwiftLint clean.

PR #109, opened 2025-09-06
Invite Deep Links

jarodl:
### Handle invite deep links by parsing `scheme://join/<code>` and `https://<domain>/join/<code>` and launching the join flow from `ConversationsView`
- Introduce `DeepLinkHandler` that validates against `ConfigManager.associatedDomain` and `ConfigManager.appUrlScheme`, and extracts invite codes from custom scheme and universal links ([DeepLinkHandler.swift](https://github.com/ephemeraHQ/convos-ios/pull/109/files#diff-262e2f8dbcd00d6d932e4f36e4ac9d7bb8d3e2cc29e15ec625ec9d57ed612a8a), [ConfigManager.swift](https://github.com/ephemeraHQ/convos-ios/pull/109/files#diff-620a0862d42f99dcebc75ce420a2c270c512437cb66cad2a713c1a28e3f8e6eb)).
- Wire `ConversationsView` to process `onOpenURL`, react to pending deep links, and present `NewConversationViewModel` with `prefilledInviteCode` ([ConversationsView.swift](https://github.com/ephemeraHQ/convos-ios/pull/109/files#diff-cd5b2e3f80b21e9829602c0b2b1b3a6c7b92d949fdb060d2f5d609050ec275e3), [NewConversationViewModel.swift](https://github.com/ephemeraHQ/convos-ios/pull/109/files#diff-9fe55a1ae8a316e2507a854b41b4ffb798098f2b5115fc2116749424cc4aa4bc)).
- Extend `NewConversationViewModel` to auto-join with a prefilled code, accept raw invite codes, surface `joinError`, bypass the scanner, and short-circuit when `DraftConversationWriterProtocol.checkIfAlreadyJoined` returns an existing conversation ID ([NewConversationViewModel.swift](https://github.com/ephemeraHQ/convos-ios/pull/109/files#diff-9fe55a1ae8a316e2507a854b41b4ffb798098f2b5115fc2116749424cc4aa4bc), [DraftConversationWriter.swift](https://github.com/ephemeraHQ/convos-ios/pull/109/files#diff-210d2935dafec930f4d82d5469bace653bd47a040e4e706580a68aaef786ec51)).
[...]

PR #1046, opened 2026-06-11
Reland "feat(agents): direct-add agent joins" (#1033)

tylerhawkes:
## Summary

Revert of the temporary revert (#1045) — the content of #1033 **plus one fix found by manual testing on the dev stack**: `ConversationMetadataWriter.addMembers` now upserts the `member` parent row before inserting `conversation_members`. A directly-added inbox (a freshly provisioned agent) has never been seen locally, so the previous insert hit `SQLite error 19: FOREIGN KEY constraint failed` — the invite flow never exercised this path because members always arrived via the stream, profile first. (Server side confirmed working in the same test: the agent provisioned, attached via the group welcome, and reached `ready` on dev.) A regression test (`DirectAddMemberPersistenceTests`) pins the fix — verified red without the upsert, green with it.

**⛔ Do not merge until the server chain is promoted to prod** (Herald → assistants worker → backend, in that order). `testflight-prod.yml` builds every push to `dev`, so merging early puts a build on the prod TestFlight track whose agent add/spawn flows 400 against the prod backend. Draft status is the merge gate; flip to ready once prod promotion is done. Note: prod Herald already has the attach endpoint (#109 merged to `main`, which deploys the prod pool — see herald-lite#111 for the dev-pool catch-up).

The dev stack runs the full chain, so dev builds exercise this flow today.

> [!NOTE]
> ### Replace slug-based agent join with direct-add by conversationId across agent and conversation flows
> - Replaces `session.requestAgentJoin(slug:)` with `session.addAgentToConversation(conversationId:)` in `AgentBuilderViewModel`, `ConversationViewModel`, and `SessionManagerProtocol`, removing all slug derivation and empty-slug guards.
> - `SessionManager.addAgentToConversation` provisions an agent via the join API, polls `getAgentJoinStatus` until an `inboxId` appears or a terminal state/timeout occurs, then adds the inbox to the conversation with retries.
> - Adds `AgentProvisionStatus` enum and `AgentJoinStatusResponse` model for typed consumption of poll responses, plus `ConvosAPIClient.getAgentJoinStatus` for the new polling endpoint.
> - Fixes a foreign-key violation in `ConversationMetadataWriter.addMembers` by inserting the parent `DBMember` row before persisting membership for never-before-seen inboxes.
> - Risk: join attempts that previously short-circuited on a missing slug now always proceed, and the full provision+poll cycle runs synchronously before the builder bundle is published.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 381a125.</sup>
> 
>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient+Models.swift
+++ b/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient+Models.swift
@@ -164,16 +164,29 @@ public enum ConvosAPI {
     // POST /v2/agents/join
     // The body is validated strictly server-side: unknown keys are rejected.
     // `templateId` is optional - omitting it requests a bare join (the
-    // backend provisions a default agent). A nil `templateId` is omitted
-    // from the encoded JSON by Codable's synthesized encoder.
+    // backend provisions a default agent). Nil optionals are omitted from
+    // the encoded JSON by Codable's synthesized encoder.
+    //
+    // Exactly one of `slug` (invite flow) or `conversationId` (direct-add).
+    // In direct-add mode the backend provisions the agent and responds with
+    // its `inboxId`; the client adds that inbox to the declared group with
+    // addMembers and is done — the runtime observes the resulting group
+    // welcome and attaches, with no confirmation call.
 
     public struct AgentJoinRequest: Codable {
-        public let slug: String
+        public let slug: String?
+        public let conversationId: String?
         public let templateId: String?
         public let options: AgentJoinOptions?
 
-        public init(slug: String, templateId: String? = nil, options: AgentJoinOptions? = nil) {
+        public init(
+            slug: String? = nil,
+            conversationId: String? = nil,
+            templateId: String? = nil,
+            options: AgentJoinOptions? = nil
+        ) {
             self.slug = slug
+            self.conversationId = conversationId
             self.templateId = templateId
             self.options = options
         }
@@ -199,10 +212,48 @@ public enum ConvosAPI {
     public struct AgentJoinResponse: Codable {
         public let success: Bool
         public let joined: Bool
+        /// Populated on every dispatch; required for the join-status poll.
+        public let instanceId: String?
+        /// Direct-add only: the agent's XMTP inbox to add with addMembers.
+        /// Nil when registration outlasted the backend's wait budget — poll
+        /// GET /v2/agents/join/:instanceId until it lands.
+        public let inboxId: String?
 
-        public init(success: Bool, joined: Bool) {
+        public init(success: Bool, joined: Bool, instanceId: String? = nil, inboxId: String? = nil) {
             self.success = success
             self.joined = joined
+            self.instanceId = instanceId
+            self.inboxId = inboxId
+        }
+    }
+
+    // MARK: - v2/agents/join/:instanceId
+
+    public struct AgentJoinStatusResponse: Codable {
+        public let success: Bool
+        public let instanceId: String
+        public let joinStatus: String
+        public let joined: Bool
+        public let inboxId: String?
+        public let conversationId: String?
+        public let joinFailureReason: String?
+
+        public init(
+            success: Bool,
+            instanceId: String,
+            joinStatus: String,
+            joined: Bool,
+            inboxId: String? = nil,
+            conversationId: String? = nil,
+            joinFailureReason: String? = nil
+        ) {
+            self.success = success
+            self.instanceId = instanceId
+            self.joinStatus = joinStatus
+            self.joined = joined
+            self.inboxId = inboxId
+            self.conversationId = conversationId
+            self.joinFailureReason = joinFailureReason
         }
     }
 
--- a/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient+Models.swift
+++ b/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient+Models.swift
@@ -255,6 +255,13 @@ public enum ConvosAPI {
             self.conversationId = conversationId
             self.joinFailureReason = joinFailureReason
         }
+
+        /// Typed view of the wire `joinStatus`. Switch over this (not the
+        /// raw string) so a terminal backend status can't be silently
+        /// missed at the call site.
+        public var provisionStatus: AgentProvisionStatus {
+            AgentProvisionStatus(wire: joinStatus)
+        }
     }
 
     // MARK: - v2/agent-templates/:id
--- a/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
+++ b/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
@@ -821,7 +821,13 @@ final class ConvosAPIClient: ConvosAPIClientProtocol, Sendable {
 
     func getAgentJoinStatus(instanceId: String) async throws -> ConvosAPI.AgentJoinStatusResponse {
         let encoded = instanceId.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? instanceId
-        let request = try authenticatedRequest(for: "v2/agents/join/\(encoded)", method: "GET")
+        var request = try authenticatedRequest(for: "v2/agents/join/\(encoded)", method: "GET")
+        // Bound a single poll: without this a hung GET stalls the caller's
+        // registration loop indefinitely, since the loop's own deadline is
+        // only checked between iterations and can't interrupt an in-flight
+        // request. Kept well under the loop's overall deadline so a stuck
+        // request fails fast and the next iteration's deadline check fires.
+        request.timeoutInterval = 10
         return try await performRequest(request)
     }
 
--- a/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
+++ b/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
@@ -83,12 +83,22 @@ public protocol ConvosAPIClientProtocol: AnyObject, Sendable {
     func renewAssetsBatch(assetKeys: [String]) async throws -> AssetRenewalResult
 
     // Agents
+    /// Exactly one of `slug` (invite flow) and `conversationId` (direct-add).
+    /// In direct-add mode the backend provisions the agent and returns its
+    /// `inboxId`; the caller adds that inbox to the declared group with
+    /// addMembers and the runtime attaches when it observes the resulting
+    /// group welcome — no further calls.
     func requestAgentJoin(
-        slug: String,
+        slug: String?,
+        conversationId: String?,
         templateId: String?,
         options: ConvosAPI.AgentJoinOptions?,
         forceErrorCode: Int?
     ) async throws -> ConvosAPI.AgentJoinResponse
+
+    /// Polls a dispatched agent's provisioning status. Direct-add callers
+    /// use it to obtain `inboxId` when the join response didn't carry one.
+    func getAgentJoinStatus(instanceId: String) async throws -> ConvosAPI.AgentJoinStatusResponse
 
     // Agent templates
     /// Public detail fetch for a published agent template, keyed by its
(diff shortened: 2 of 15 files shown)


ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift
```diff
@@ -1131,3 +1109,58 @@ extension SessionManager {
...
+            throw APIError.agentPoolTimeout
+        }
+
+        // 2. Add: a plain member add by this device — synchronous, no DM
+        //    round-trip, no online-accepter dependency. The add durably
+        //    publishes a group welcome for the agent's inbox; the runtime
+        //    observes it and attaches, so the agent boots even if this app
+        //    is killed on the next line.
+        try await messagingService().conversationMetadataWriter()
+            .addMembers([agentInboxId], to: conversationId)
```
lourou: We'd end up in an unrecoverable state after provision succeeds but addMembers fails.

By this point we hold `instanceId` and `inboxId` from a successful provision POST. If this `addMembers` call throws or is cancelled (transient MLS error, key package not yet published, network drop, local DB write failure), both ids are discarded - callers retain only `templateId`.

A retry re-POSTs `agents/join` and provisions a *new* agent instance (if not idempotent on the backend?), `instanceId` is not logged here, so backend cleanup is impossible.

Fix: persist/return `instanceId` and `inboxId` before attempting the add; on add failure, retry the MLS add for the same `inboxId` rather than re-provisioning; on terminal failure make a best-effort backend release call with the persisted `instanceId`. At minimum, log `instanceId` on every failure path.

ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift
```diff
@@ -1131,3 +1109,58 @@ extension SessionManager {
...
+
+        var inboxId = response.inboxId
+        let deadline = Date().addingTimeInterval(Self.agentRegistrationDeadline)
+        while inboxId == nil {
+            guard Date() < deadline else { throw APIError.agentPoolTimeout }
+            try await Task.sleep(nanoseconds: UInt64(Self.agentRegistrationPollInterval * 1_000_000_000))
+            let status = try await apiClient.getAgentJoinStatus(instanceId: instanceId)
+            guard status.joinStatus != "failed" else {
+                throw APIError.agentProvisionFailed
+            }
```
lourou: Only `"failed"` is handled, but `AgentJoinStatus` also defines `noAgentsAvailable`. If the backend emits it, the loop spins to the 30s deadline and throws `agentPoolTimeout` so the user sees "timed out" instead of "no agents available".

Fix: treat `"no_agents_available"` as terminal too; better, decode `joinStatus` as the `AgentJoinStatus` enum and `switch` exhaustively so new statuses can't be silently missed.

There should always be agents available, but...!

ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift
```diff
@@ -1131,3 +1109,58 @@ extension SessionManager {
...
+            inboxId = status.inboxId
+        }
```
lourou: The poll exits on `inboxId != nil`, not on a positive ready/terminal status. If the backend returns `inboxId` before the agent's key package is published, `group.addMembers` will throw (no key packages) or add a member that never decodes the welcome - the local DB records membership while the agent silently never joins.

Question re: the backend implementation: does `inboxId` in the poll response imply the key package is already published and ready for `addMembers`? If not, gate the exit on a positive ready/completed status rather than mere `inboxId` presence. The test stubs returning `inboxId` while `joinStatus == "starting"` model a combination that should never occur in the real protocol.

ConvosCore/Sources/ConvosCore/API/ConvosAPIClient+Models.swift
```diff
@@ -199,10 +212,48 @@ public enum ConvosAPI {
...
+            self.inboxId = inboxId
+        }
+    }
+
+    // MARK: - v2/agents/join/:instanceId
+
+    public struct AgentJoinStatusResponse: Codable {
+        public let success: Bool
+        public let instanceId: String
+        public let joinStatus: String
```
lourou, line 235: `joinStatus` is declared as a raw `String` rather than the existing `AgentJoinStatus` enum. That is what makes it easy to miss terminal statuses like `no_agents_available` at the call site in `SessionManager.addAgentToConversation`. Decoding into the enum and switching exhaustively would make the missing-status bug a compile error instead of a silent spin-to-timeout.

This is the root cause of the terminal-status gap

ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
```diff
@@ -807,6 +819,12 @@ final class ConvosAPIClient: ConvosAPIClientProtocol, Sendable {
         }
     }
 
+    func getAgentJoinStatus(instanceId: String) async throws -> ConvosAPI.AgentJoinStatusResponse {
+        let encoded = instanceId.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? instanceId
+        let request = try authenticatedRequest(for: "v2/agents/join/\(encoded)", method: "GET")
+        return try await performRequest(request)
```
lourou, line 831: The 30s timeout isn't actually enforced.

The loop only checks the deadline between iterations, it can't interrupt an in-flight request, and `getAgentJoinStatus` has no timeout, so a single hung GET stalls the loop indefinitely.

Simplest fix: give the GET a short `timeoutInterval` (like the provision POST already has) so a stuck request fails fast and the loop's own deadline check can fire.

ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift
```diff
@@ -1131,3 +1109,58 @@ extension SessionManager {
...
+        //    returns it. No invite slug — this device performs the add, and
+        //    the declared conversationId tells the runtime which group to
+        //    watch for the resulting welcome.
+        let response = try await apiClient.requestAgentJoin(
+            slug: nil,
+            conversationId: conversationId.lowercased(),
+            templateId: templateId,
+            options: options,
+            forceErrorCode: forceErrorCode
+        )
```
lourou: Is a repeat `agents/join` POST for the same `conversationId`+`templateId` idempotent (returns the existing instance), or does it spawn a new agent every time? The answer decides whether the simpler client fix is sufficient.

We'd want to avoid having two agents joining the same convo.

ConvosCore/Sources/ConvosCore/Sessions/SessionManager.swift
```diff
@@ -1131,3 +1109,58 @@ extension SessionManager {
...
+            try await Task.sleep(nanoseconds: UInt64(Self.agentRegistrationPollInterval * 1_000_000_000))
+            let status = try await apiClient.getAgentJoinStatus(instanceId: instanceId)
+            guard status.joinStatus != "failed" else {
+                throw APIError.agentProvisionFailed
+            }
+            inboxId = status.inboxId
+        }
+        guard let agentInboxId = inboxId else {
+            throw APIError.agentPoolTimeout
+        }
```
lourou: This guard's `else` branch is unreachable.

The `while inboxId == nil` loop only exits when `inboxId` is non-nil, so `agentPoolTimeout` here can never fire.

ConvosCore/Sources/ConvosCore/API/MockAPIClient.swift
```diff
@@ -96,12 +96,23 @@ final class MockAPIClient: ConvosAPIClientProtocol, Sendable {
...
+        .init(success: true, joined: true, instanceId: "mock-instance", inboxId: "mock-agent-inbox")
+    }
+
+    func getAgentJoinStatus(instanceId: String) async throws -> ConvosAPI.AgentJoinStatusResponse {
+        .init(
+            success: true,
+            instanceId: instanceId,
+            joinStatus: "starting",
+            joined: false,
+            inboxId: "mock-agent-inbox"
```
lourou, line 117: Nickpick here: but this returns `joinStatus: "starting"` together with a non-nil `inboxId`, a combination that cannot occur in the real protocol and which masks the poll loop entirely (the `while inboxId == nil` loop never iterates, so `getAgentJoinStatus` is never exercised).

Use `joinStatus: "completed"` with a non-nil inbox, or `joinStatus: "starting"` with `inboxId: nil`, so the stub is protocol-faithful.

ConvosCore/Tests/ConvosCoreTests/TestStubAPIClientDefaults.swift
```diff
@@ -36,12 +36,31 @@ extension ConvosAPIClientProtocol {
...
+    /// Default for the direct-add status poll so pre-existing stubs don't
+    /// re-stub it. Tests that exercise the direct-add flow specifically
+    /// should override this on their fixture.
+    func getAgentJoinStatus(instanceId: String) async throws -> ConvosAPI.AgentJoinStatusResponse {
+        ConvosAPI.AgentJoinStatusResponse(
+            success: true,
+            instanceId: instanceId,
+            joinStatus: "starting",
+            joined: false,
+            inboxId: "test-agent-inbox"
```
lourou, line 64: Same contradictory state as the mock: `joinStatus: "starting"` with a non-nil `inboxId` can't happen in the real protocol, and it means every test using this default skips the poll loop (the `while inboxId == nil` condition is false immediately, so `getAgentJoinStatus` is never called).

Pair `"completed"` with a non-nil inbox, or `"starting"` with `nil`. See T1: the entire polling loop and all its branches are currently untested.

lourou (changes requested):
Reland looks good! Review has a few unhappy-path asks (provision/addMembers state, recovery, terminal-status) and a flag to confirm prod backend promotion before merge.

tylerhawkes:
Thanks for the thorough pass @lourou — all addressed in `381a1253`, kept as a separate commit on top of the reland so the delta is easy to re-review.

**Type-safe status (root cause: `joinStatus: String`)** — added `AgentProvisionStatus` with a `provisionStatus` accessor on `AgentJoinStatusResponse`; the poll now `switch`es over it exhaustively. One deliberate call: I kept the *wire* field a lenient `String` decoded into the enum via an `.unknown(String)` fallback, rather than decoding the field directly as the enum. A hard enum decode would **throw** and break the poll the first time the backend adds a status this build doesn't know — strictly worse than a handled `.unknown`. The compile-time safety you wanted lives at the `switch`: modeled cases are exhaustive; anything new surfaces as `.unknown` and is logged, never silently treated as "keep waiting".

**`no_agents_available` (1150)** — now modeled and terminal → throws `.noAgentsAvailable`, so the user sees "no agents available" instead of spinning to "timed out". (In practice it's a provision-time 503 already mapped in `requestAgentJoin`; modeling it on the poll too is belt-and-suspenders.)

**Exit gated on `inboxId`, not a ready status (1152)** — this is the backend contract, not an accident: in direct-add, `pollUntilRegistered` (backend `join.ts`) sets `inboxId` **only after** the agent's identity is registered and key packages are published, so `inboxId` *is* the readiness signal and a subsequent `addMembers` has key packages to consume. One correction on the premise though: `{joinStatus: "starting", inboxId: <set>}` is a **real** direct-add state — the status stays `starting` until the agent actually *joins*, which only happens after our `addMembers`. So the inbox check is the correct exit; I've documented why and the loop detects failure via the status `switch`.

**30s timeout not enforced (825)** — `getAgentJoinStatus` now sets `request.timeoutInterval = 10` (well under the loop's 30s deadline) so a hung GET fails fast and the loop's own deadline check can fire.

**Unreachable guard (1155)** — gone; the inbox binds at the single exit point of the extracted helper.

**Unrecoverable after provision-ok / add-fails (1163) + idempotency (1137)** — confirmed the backend is **not** idempotent per `conversationId` (no dedup in `join.ts`; each POST spawns a fresh instance), so re-provisioning on add-failure would orphan the first instance and risk two agents in one conversation. Fix: on `addMembers` failure we **retry the same `inboxId`** (bounded, 3×) and never re-provision, and `instanceId` + `inboxId` are logged on every failure path. An add that exhausts its retries leaves a provisioned-but-unadded instance, which the runtime reaps on its own attach timeout (~10 min); there's no client-callable release endpoint today, so the logged `instanceId` is the correlation hook. I can add a backend release/cancel call as a follow-up if you'd like one wired up.

**Contradictory stubs (Mock:114, Stub:62)** — made coherent (`joined` ⇒ inbox present). The masking you spotted actually originated in the *provision* stub returning a non-nil `inboxId` (so `while inboxId == nil` never iterated). The poll loop and every branch now have direct coverage in `DirectAddProvisionPollTests` — fast-path, poll-until-registered, `failed`, `no_agents_available`, unknown-status, deadline timeout, missing-`instanceId` — with injected `now`/`sleep` so they run with no real waiting.

On the prod-backend flag: Nick has confirmed that everything went out last Friday.

lourou (approved):
Solid!

PR #1065, opened 2026-06-12
fix(invites): dedupe join requests by message ID so removing a member sticks

yewreeka:
## Problem

Removing an agent (or any invited member) from a conversation could re-add them within seconds.

Join-request DMs are durable, and several paths revalidate them after the member list changes: the live message stream, batch catch-up (`runInviteBatch`, up to 24h lookback), the 150s agent-join poll, and the notification service extension. The only dedupe in `InviteCoordinator.addJoinerToGroup` was **"is the joiner currently a member"** - so the moment a member was removed, their already-honored join request looked actionable again, and the next revalidation pass silently called `addMembers` with the still-valid invite. For agents this presented as the agent instantly rejoining after removal (the convos-assistants runtime never re-requests - herald-lite sends one join request and explicitly does not retry; the replay was entirely creator-side).

## Fix

A **handled-request ledger keyed by message ID**: a join-request message admits its sender at most once.

- `ConvosInvites`: new `HandledJoinRequestStoreProtocol` (+ in-memory default). `InviteCoordinator` consults the ledger before adding, marks the message after a successful add, and also marks when a pass observes the joiner already in the group (retires the herald typed+text duplicate and requests honored before the ledger existed).
- `processDm` skips ledger-handled candidates without ending the scan, so a fresh re-join request in the same DM still gets processed.
- `ConvosCore`: `DatabaseHandledJoinRequestStore` persists the ledger in the shared app-group database (app + NSE, fresh manager per batch), with 30-day pruning. Wired as the default in `InviteJoinRequestsManager`.

**Removal is not a block**: a removed member can rejoin with the same valid invite by sending a fresh join request (new message ID).

## Tests

- New integration test (local XMTP node): honored request replayed after removal stays inert; a fresh request from the removed member is honored again.
- Unit tests for the GRDB store (read-back, idempotent mark, retention pruning).
- Full suites pass: ConvosCore 1260 tests / ConvosInvites 170 tests. SwiftLint clean.

## Possible follow-ups (not in this PR)

- A second device/installation of the same inbox keeps its own ledger; an in-band handled marker or invite-tag rotation would extend the guarantee across installations.
- Notify convos-assistants on removal (herald exposes `leaveConversation`/`deleteAgent`) so removed agent instances tear down instead of lingering.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1065"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783870380&installation_id=128169237&pr_number=1065&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1065&signature=2dca80c69ee69dc78b0db16d70c63341943046108e49e5af3883fb4d57db367d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Dedupe join requests by message ID to prevent removed members from being re-added
> - Adds a `handledJoinRequest` DB table and `DatabaseHandledJoinRequestStore` to persist handled join-request message IDs with a 30-day retention window, so deduplication survives app restarts.
> - Introduces a new `invite_join_handled` content type and `InviteJoinHandledCodec`; after honoring a join request, the creator posts a marker to the DM so other installations can skip reprocessing.
> - `InviteCoordinator.processJoinRequest` now checks the ledger first and treats already-handled message IDs as `alreadyMember` without repeating crypto work or sending duplicate error replies.
> - The DM scan loop no longer exits on the first `alreadyMember` result, continuing to find actionable requests while still reporting `alreadyMember` if nothing else applies.
> - Handled markers are only sent when the joiner has demonstrated typed-client capability, preserving compatibility with text-only (pre-typed) clients.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f58aec7.</sup>
> 
>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
+++ b/ConvosCore/Sources/ConvosCore/Inboxes/SessionStateMachine.swift
@@ -1114,6 +1114,7 @@ public actor SessionStateMachine: SessionStateManagerProtocol {
                 GroupUpdatedCodec(),
                 ExplodeSettingsCodec(),
                 InviteJoinErrorCodec(),
+                InviteJoinHandledCodec(),
                 ProfileUpdateCodec(),
                 ProfileSnapshotCodec(),
                 JoinRequestCodec(),
--- a/ConvosCore/Sources/ConvosCore/Storage/Database Models/DBHandledJoinRequest.swift
+++ b/ConvosCore/Sources/ConvosCore/Storage/Database Models/DBHandledJoinRequest.swift
@@ -1,0 +1,16 @@
+import Foundation
+import GRDB
+
+/// A join-request message that already admitted its sender to a group.
+/// Keyed by XMTP message ID; see `DatabaseHandledJoinRequestStore`.
+struct DBHandledJoinRequest: FetchableRecord, PersistableRecord, Codable, Hashable {
+    static let databaseTableName: String = "handledJoinRequest"
+
+    enum Columns {
+        static let messageId: Column = Column(CodingKeys.messageId)
+        static let handledAt: Column = Column(CodingKeys.handledAt)
+    }
+
+    let messageId: String
+    let handledAt: Date
+}
--- a/ConvosCore/Sources/ConvosCore/Storage/SharedDatabaseMigrator.swift
+++ b/ConvosCore/Sources/ConvosCore/Storage/SharedDatabaseMigrator.swift
@@ -207,7 +207,20 @@ extension SharedDatabaseMigrator {
 
         migrator.registerMigration("addConnectionGrantBundleScope", migrate: Self.addConnectionGrantBundleScope)
 
+        migrator.registerMigration("createHandledJoinRequest", migrate: Self.createHandledJoinRequest)
+
         return migrator
+    }
+
+    /// Ledger of join-request message IDs that already admitted their
+    /// sender, shared with the notification service extension. Keyed by
+    /// message ID so revalidation passes cannot re-add a member the user
+    /// removed; see `DatabaseHandledJoinRequestStore`.
+    private static func createHandledJoinRequest(_ db: Database) throws {
+        try db.create(table: "handledJoinRequest") { t in
+            t.column("messageId", .text).notNull().primaryKey()
+            t.column("handledAt", .datetime).notNull()
+        }
     }
 
     /// Persisted "the local user was removed from this conversation" marker.
--- a/ConvosCore/Sources/ConvosCore/Storage/SharedDatabaseMigrator.swift
+++ b/ConvosCore/Sources/ConvosCore/Storage/SharedDatabaseMigrator.swift
@@ -221,6 +221,11 @@ extension SharedDatabaseMigrator {
             t.column("messageId", .text).notNull().primaryKey()
             t.column("handledAt", .datetime).notNull()
         }
+        try db.create(
+            index: "handledJoinRequest_handledAt",
+            on: "handledJoinRequest",
+            columns: ["handledAt"]
+        )
     }
 
     /// Persisted "the local user was removed from this conversation" marker.
--- a/ConvosCore/Sources/ConvosCore/Syncing/DatabaseHandledJoinRequestStore.swift
+++ b/ConvosCore/Sources/ConvosCore/Syncing/DatabaseHandledJoinRequestStore.swift
@@ -28,11 +28,13 @@ final class DatabaseHandledJoinRequestStore: HandledJoinRequestStoreProtocol, Se
     }
 
     func markHandled(messageId: String) async {
+        let now = Date()
+        let retentionCutoff = now.addingTimeInterval(-Constant.retention)
         do {
             try await databaseWriter.write { db in
-                try DBHandledJoinRequest(messageId: messageId, handledAt: Date()).save(db)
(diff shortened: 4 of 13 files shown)


yewreeka:
Addressed findings from two reviews (the Claude workflow review above, plus a second adversarial review pass) in 1a23fe5c:

**Blocker (second review): marker broke the joiner-side consent bump.** `hasOutgoingJoinRequest` required the joiner's request to be the DM's *literal last* message. The new creator-sent marker displaced it the moment a join was honored, so a joiner that synced the DM before processing the group welcome would never get the `.unknown -> .allowed` consent bump and the joined conversation could stay hidden with no self-heal path. `lastMessageAsSignedInvite` now scans a small recent window for the joiner's own most recent invite. Regression-tested in the marker integration test (joiner-side `hasOutgoingJoinRequest` asserted true after the marker lands).

**Should-fix (second review): newest-50 marker scan window.** A chatty DM (~13+ herald attempts) could push a marker out of the window while its request was still inside the 24h catch-up window, letting a cross-installation replay through. Marker fetches are now anchored at the covered request's send time (`afterNs`) - precise and naturally bounded, since a covering marker is always sent after its request. The error-reply scan keeps its bounded window (worst case there is one duplicate reply, now documented on the constant).

**From the Claude review:**
- Added index `handledJoinRequest_handledAt` for the retention prune (edited the unshipped migration in place).
- Prune timestamps computed once outside the write transaction.
- Documented the scan-limit failure mode on the constant.
- Prune test already asserts recent rows survive (`msg-new` check) - no change needed.
- Prune rate-limiting: deferred; the table is tiny (30-day retention, join requests are rare) and the index makes the per-write prune cheap.

Also moved the ledger check ahead of all validation, so replayed requests skip signature/decrypt work and can't emit stale error replies when group state changed since the original accept.

Full suites green after the changes: ConvosCore 1261, ConvosInvites 177.

yewreeka:
**Cross-version compatibility hardening** (bfb21611): a 2.0.1 creator sending the handled marker to a **2.0.0 joiner** would race that joiner's `hasOutgoingJoinRequest` check, which on pre-typed builds requires the join request to be the DM's *literal last* message - the marker could displace it before the welcome is processed, leaving the joined conversation hidden with no consent bump. We can fix our own check (previous commit) but not shipped 2.0.0 code.

The marker is now **gated on the joiner being typed-capable**: it's only sent when the DM contains a typed join request from the joiner. Typed joiner sends landed post-2.0.0 (#1034) and ship in the same release as the windowed check, so a typed request proves the joiner tolerates creator bookkeeping in the DM. The gate checks the whole DM rather than the honored message because herald's *text* copy is newer and usually the one honored - agents keep full marker protection either way (herald never inspects DM last-messages; it diffs the conversation list).

Trade-off: joins from 2.0.0 text-only clients get no marker, so no cross-installation dedupe for those specific joins (the same-device ledger still covers them) - their wire experience is byte-identical to today. Gap closes as joiners update; text acceptance removal later retires the gate naturally.

Tests: text-only join asserts zero markers; the typed+text pair asserts exactly one and cross-install dedupe still holds. Full suites green (ConvosCore 1261 / ConvosInvites 177).

yewreeka:
**Second adversarial review pass** - verdict was *safe to merge, no blockers* (it independently re-verified the herald typed+text flow, `afterNs`/sort semantics against the SDK source, marker forgery resistance, the 2.0.0 joiner paths, NSE wiring, and migration soundness). Two narrow should-fixes landed in f58aec7a rather than as follow-ups:

1. **Unverified ledger writes** (security posture): `processDm`'s marker pre-filter parsed slugs without signature verification and persisted them to the ledger - a tampered slug predating a same-tag marker could permanently evade `denyDmForMaliciousInvite` (DM denial + spammer block). Marker suppression now lives only behind signature verification in `addJoinerToGroup`; `processDm` handles `.alreadyMember` as collect-and-continue instead of scan-ending, which also stops a suppressed candidate from starving a pending request for a second group in the same DM (the review's finding 6).

2. **Pre-ledger joins were invisible cross-installation**: the membership-dedupe branch only backfilled the local ledger, so joins honored before this ships (or whose best-effort marker send failed) had a 24h replay window on other installations. That branch now also sends the marker, behind the same typed-capability gate - reaching it means no covering marker exists, so no extra dedupe is needed.

Also added the review's suggested test-honesty case: a text-only joiner exercising the **ledger in isolation** (gate suppresses the marker; assert zero markers, replay inert, fresh text request rejoins).

Deferred as polish (per review): exact `sentAtNs` threading instead of the Date->ns roundtrip (~256ns error vs millisecond-scale margins), and the pre-existing app/NSE check-then-act race whose worst case is a duplicate no-op add.

Suites green: ConvosCore 1262 / ConvosInvites 177.

PR #1087, opened 2026-06-17
feat(agents): publish user timezone to agents

lourou:
## What

Implements the **producer (iOS) side** of giving a Convos AI agent the user's IANA time zone, invisibly, so it can answer "what time is it" / "remind me at 9am" correctly and handle DST. Nothing is rendered in chat. This is the iOS half only; the backend (`v2/agents/join`) and the agent runtime (Hermes) consume these values in a separate, coordinated change.

The same wire name `timezone` is used on both channels; the role distinction lives in comments, not the field name (a source-agnostic name is preferred over `creatorTimezone` because the assistant only needs the value).

## Two channels

**Channel A — join-request field (creator baseline).** Adds `timezone: String?` to `ConvosAPI.AgentJoinRequest` and populates it at the single construction site. The value is captured on the main actor before any async hop (`await MainActor.run { TimeZone.current.identifier }` in `SessionManager.addAgentToConversation`) and threaded through `ConvosAPIClientProtocol.requestAgentJoin`. It carries the conversation creator's device tz as the agent's baseline/default; set once at join time (does not track travel/DST). Optional and nil-omitted by Codable, so the wire format stays backward-compatible until the backend reads it.

**Channel B — per-sender `ProfileUpdate.metadata` (ongoing per-user).** Publishes the current user's own tz under metadata key `"timezone"`, scoped to conversations that contain an agent member. Triggered:
- right after a successful agent join (`SessionManager.addAgentToConversation`), and
- opportunistically on app foreground (`ConvosApp` scenePhase `.active`), after a 5s settle delay, throttled to publish only when the current tz differs from the last-published value (persisted per-conversation in the app-group `UserDefaults`), once per foreground session (re-armed on `.background`).

Both triggers gate on the conversation actually having an agent member, and only ever run from the foreground main app.

## The shared serialization choke point

The per-sender metadata map is rewritten whole on each publish (read existing -> merge key -> publish). A connections write and a timezone write could interleave on that non-atomic sequence and clobber each other's key — silent data loss. New `ProfileMetadataWriter` (`@MainActor public final class`) owns the single choke point: read existing metadata, apply a caller-supplied closure, publish the merge. An internal serial task chain makes each `updateMetadata` atomic across callers even across `await` suspension points. `CloudConnectionGrantWriter` is refactored to route its `connections` write through this shared instance (behavior unchanged; all 23 existing connection-grant tests pass), and the timezone publish uses it for the `timezone` key, so the two can never race.

## Tests

- New `ProfileMetadataWriterTests` (3): merge-on-existing, empty-map-publishes-nil, concurrent-writes-preserve-both-keys.
- `ConnectionGrantWriterTests` updated for the new constructor; all 23 pass — no regression.

## Verification

- `swift build --package-path ConvosCore` — succeeds.
- `xcodebuild build` `Convos (Dev)` on iPhone 17 simulator — succeeds, zero errors, zero long-type-check warnings.
- `swiftlint --strict` on changed files — clean.

## Notes

Producer half only. The backend must accept the join `timezone` field and the runtime must read `metadata["timezone"]` for it to have any effect; until then iOS sends the bytes and they are ignored.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1087"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784308248&installation_id=128169237&pr_number=1087&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1087&signature=0f729eea769a684f0c5855f2b80b6c9ca2be25853469a7317710604e66d3152c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Publish user timezone to agent conversations on join and app foreground
> - Adds the device timezone to `requestAgentJoin` API calls and writes it into per-sender `ProfileUpdate.metadata` after an agent is added to a conversation.
> - Introduces `AgentTimezonePublisher` to publish and throttled-republish timezone metadata across all agent conversations for the current inbox.
> - On each app foreground, schedules a one-time-per-foreground republish of agent timezones after a 5-second delay via `ForegroundOnceGuard` and `session.republishAgentTimezones()`.
> - Adds `ProfileMetadataWriter` to serialize concurrent `ProfileUpdate.metadata` writes and prevent key clobbering between operations like timezone and connection grant updates.
> - Risk: timezone publish failures are best-effort and do not fail the agent join, so agents may temporarily have stale or missing timezone metadata.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e876db.</sup>
> 
>

### DIFF
--- a/Convos/ConvosApp.swift
+++ b/Convos/ConvosApp.swift
@@ -15,6 +15,11 @@ struct ConvosApp: App {
     let coreActions: any CoreActions
     let conversationsViewModel: ConversationsViewModel
     let profileSettingsViewModel: ProfileSettingsViewModel = .shared
+    /// Guards the opportunistic agent-timezone republish to once per foreground
+    /// session, so background-foregrounding the app repeatedly does not re-run
+    /// it. A reference type so the flag survives across `onChange` invocations
+    /// of the value-type App.
+    private let timezoneForegroundGuard: ForegroundOnceGuard = ForegroundOnceGuard()
 
     init() {
         FileDescriptorDiagnostics.raiseSoftLimit(to: 512)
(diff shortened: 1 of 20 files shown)


cameronvoell (approved):
excited for when this is wired up with the backend. will be a big improvement over the current state of the agent always having the same first question. 🚀

PR #1168, opened 2026-07-13
Fix main-thread hangs reported by Sentry app-hang monitoring

yewreeka:
## Summary

Fixes the four Sentry app-hang issues (production, "App hanging for at least 2000 ms") with definitively identified causes. Each was a main-thread stall captured with a full stack:

| Sentry issue | Root cause | Fix |
|---|---|---|
| [CONVOS-IOS-3T](https://ephemerahq.sentry.io/issues/?query=CONVOS-IOS-3T) | `ContactsViewModel.init` ran a synchronous `fetchAll()` during SwiftUI body evaluation and stalled on the GRDB reader-pool semaphore | Initial fetch runs detached; publisher-delivered data wins races. `rebuildSections()` also stops doing a synchronous `sourceConversations` read per keystroke — it reads an off-main-refreshed cache |
| [CONVOS-IOS-3Y](https://ephemerahq.sentry.io/issues/?query=CONVOS-IOS-3Y) | `CreditsRepository.balancePublisher` used `.immediate` scheduling → synchronous WAL-snapshot read at subscription time, subscribed during layout via `.onReceive` (MainTabView, ConversationMemberView) | Default async scheduler; all consumers just assign state |
| [CONVOS-IOS-3Q](https://ephemerahq.sentry.io/issues/?query=CONVOS-IOS-3Q) | Conversation-list rows resolve member names via `contact(for:)` — a DB point read **per member per row** mid-body (`ConversationsListItem.body` → `Conversation.computedDisplayName`) | `contact(for:)` is now a protocol requirement; the live repository answers from an in-memory contact-table mirror (lazily started observation, point-read fallback until warm). `MessagingService.contactsRepository()` returns a shared instance so the cache is reused across body evaluations |
| [CONVOS-IOS-2Q](https://ephemerahq.sentry.io/issues/?query=CONVOS-IOS-2Q) | `BackgroundUploadManager` created its background `URLSession` inline in `ConvosApp.init`; session creation blocks on a synchronous XPC handshake with `nsurlsessiond` | Session creation happens on a serial background queue; accessors rendezvous via `sync` (a no-op once created) |

Underlying all the GRDB hangs: the waits were on `Pool.get` — the reader pool was exhausted by long-running reads elsewhere, so even cheap main-thread reads stalled for seconds. This PR removes the main-thread reads; auditing what holds readers that long is a follow-up.

Not addressed (separate follow-ups): TextKit2 layout stall on a pathological message (CONVOS-IOS-2H/2M/3R — needs a product decision on truncation), SwiftUI body cost in ConversationView (3S), CoreUI glyph-catalog I/O stall (28).

## Test plan

- [ ] CI build + unit tests (local build blocked by disk space on the dev machine)
- [ ] Manual: Contacts tab first open shows list; search filtering works; "met in" subtitles populate
- [ ] Manual: credits badge appears in MainTabView; out-of-credits state in contact sheet
- [ ] Manual: conversation list renders member names; contact rename propagates
- [ ] Manual: photo send via background upload completes; upload resumes when app relaunched in background

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1168"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786556618&installation_id=128169237&pr_number=1168&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1168&signature=7480d727e6d14ef151397594303c5e5d087a5d0d1ff94420f49bc523df345fce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

> [!NOTE]
> ### Fix main-thread hangs in contacts, credits, and background upload initialization
> - Moves `ContactsViewModel` source-conversation lookups off the main thread using an in-memory cache populated by a `ValueObservation`, with a generation token to discard stale results; rebuilds now read from cache instead of hitting the database synchronously.
> - Adds an in-memory cache to `ContactsRepository.contact(for:)` backed by a lazy `ValueObservation`, falling back to a point read until the first observation delivers.
> - Defers `BackgroundUploadManager` background `URLSession` creation to an async task on a serial queue, eliminating a synchronous XPC handshake on the main thread at init.
> - Removes `.immediate` scheduling from `CreditsRepository.balancePublisher` so the initial value is delivered asynchronously rather than via a blocking main-thread read.
> - Behavioral Change: `CreditsRepository` subscribers and `ContactsViewModel` subtitles now receive their first value asynchronously instead of synchronously at subscription time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7a8ecc0.</sup>
> 
>

### DIFF
--- a/Convos/Contacts/ContactsViewModel.swift
+++ b/Convos/Contacts/ContactsViewModel.swift
@@ -72,6 +72,9 @@ final class ContactsViewModel {
     /// contact set changes; `rebuildSections()` reads only this cache so
     /// keystroke/filter changes never touch the database mid-render.
     private var sourceConversationsCache: [String: ContactSourceConversation] = [:]
+    /// Monotonic token for in-flight source-conversation refreshes, so a
+    /// slow older fetch can't overwrite the result of a newer one.
+    private var sourceConversationsGeneration: Int = 0
 
     /// Shared suggested-agents fetch/pagination state. A nil service yields no
     /// section (e.g. previews that don't wire one).
@@ -129,10 +132,14 @@ final class ContactsViewModel {
     /// so subtitles fill in as soon as the fetch lands.
     private func refreshSourceConversations() {
         let ids = Set(allContacts.compactMap { $0.addedViaConversationId })
+        sourceConversationsGeneration += 1
+        let generation = sourceConversationsGeneration
         Task.detached(priority: .userInitiated) { [weak self, contactsRepository = self.contactsRepository] in
             guard let sources = try? contactsRepository.sourceConversations(forIds: ids) else { return }
             await MainActor.run { [weak self] in
-                guard let self, self.sourceConversationsCache != sources else { return }
+                guard let self,
+                      generation == self.sourceConversationsGeneration,
+                      self.sourceConversationsCache != sources else { return }
                 self.sourceConversationsCache = sources
                 self.rebuildSections()
             }
--- a/Convos/Contacts/ContactsViewModel.swift
+++ b/Convos/Contacts/ContactsViewModel.swift
@@ -64,6 +64,14 @@ final class ContactsViewModel {
     private let contactsRepository: any ContactsRepositoryProtocol
     private var cancellable: AnyCancellable?
     private var allContacts: [Contact] = []
+    /// True once the repository publisher has delivered a value; gates the
+    /// best-effort initial fetch so it can't overwrite fresher data.
+    private var hasReceivedContacts: Bool = false
+    /// Source-conversation metadata for the "you met them in X" subtitles,
+    /// keyed by conversation id. Refreshed off the main thread when the
+    /// contact set changes; `rebuildSections()` reads only this cache so
+    /// keystroke/filter changes never touch the database mid-render.
+    private var sourceConversationsCache: [String: ContactSourceConversation] = [:]
 
     /// Shared suggested-agents fetch/pagination state. A nil service yields no
     /// section (e.g. previews that don't wire one).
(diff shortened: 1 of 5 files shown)


yewreeka:
Addressed the review in cbd072e4. Disposition of each finding:

**Taken:**
- **#3 (session() deadlock guard):** added `dispatchPrecondition(condition: .notOnQueue(sessionQueue))` — no current caller is on that queue, but the assertion makes the invariant explicit.
- **#4 (cache observation error recovery):** an observation error now clears `cacheObservationStarted`, so the next `contact(for:)` call restarts the observation instead of falling back to point reads forever.
- **#5 (stale sourceConversations):** added a generation token so a slow older fetch can't overwrite a newer one.

**Not taken (with reasons):**
- **#1 (TOCTOU in ContactsViewModel):** not a reachable race — the publisher sink (`receive(on: DispatchQueue.main)`) and the `MainActor.run` block both execute serially on the main thread, and there is no suspension point between the `hasReceivedContacts` check and `applyContacts` inside the block. The suggested guard inside `applyContacts` would actually break every publisher delivery after the first (the sink sets the flag before calling it).
- **#2 (double-start of cache observation):** the code already sets `cacheObservationStarted = true` inside the lock before unlocking — exactly the pattern the review proposes; only one caller can ever observe `shouldStart == true`.
- **#6 (lazy var for backgroundSession):** a `lazy var` initializes on the first caller's thread, which can be the main thread — that reintroduces the launch hang (CONVOS-IOS-2Q) this PR fixes. The queue-owned IUO is deliberate.
- **#7 (scheduling comment):** GRDB's `publisher(in:scheduling:)` defaults to `.async(onQueue: .main)`, not `.immediate` — the comment is accurate as written.
- **#8 (search debounce):** agreed as a possible follow-up; out of scope here since the DB read is gone from the keystroke path.

yewreeka:
Simulator smoke test of the four touched flows (fresh sim, Dev scheme, fresh account):

- **Launch (2Q path):** app cold-launches cleanly through `ConvosApp.init` with the off-main background-session creation; onboarding completes.
- **Contacts tab (3T path):** list renders with server-paged suggested agents; keystroke search filters correctly and shows the proper filtered empty state; agent contact card opens.
- **Conversations list (3Q path):** renders with title/preview resolution through the cache-backed resolver.
- **Photo send (2Q upload path):** end-to-end success — compress → presigned URL → S3 upload → `sent_message is_success: true, attachment_types: ["image/jpeg"]`, local attachment state migrated to remote digest.
- **Credits (3Y path):** MainTabView subscriptions active with async-scheduled balance observation; no stall at layout.

Caveat: a fresh account can't reproduce the original reader-pool starvation, so this verifies no regression in the refactored flows rather than re-observing the hangs.

PR #1293, opened 2026-08-06
default true for the participation control in production

humanagent:
## What

Listen (the per-conversation agent participation control: Speak freely / Listen mode / Paused) defaulted on only for the non-production builds. On production the control was simply absent until someone unlocked the prod debug menu and flipped it by hand.

Default it on everywhere, production included.

## How

One line: `listenParticipationDefault` goes from `!isProduction` to `true`.

Everything else about the flag is unchanged. The stored UserDefaults value still takes precedence, so an explicit toggle in either direction is remembered and wins over the default - turning it off from the prod debug menu still sticks. The prod debug menu row stays where it is.

## Testing

- `swiftlint --strict` clean.
- `xcodebuild build -scheme "Convos (Dev)" -configuration Dev` succeeds.
- No behavior change for anyone who has already toggled the flag; this only moves the first-run default on production builds.

> [!NOTE]
> ### Enable listen participation feature flag by default in all environments
> Changes `FeatureFlags.listenParticipationDefault` from a computed var that returned `false` in production to a `static let` hardcoded to `true`, removing the dependency on `ConfigManager`. Behavioral Change: users in production with no `UserDefaults` override will now have listen participation enabled by default.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6b24869.</sup>
> 
>

### DIFF
--- a/Convos/Config/FeatureFlags.swift
+++ b/Convos/Config/FeatureFlags.swift
@@ -71,10 +71,10 @@ final class FeatureFlags {
     /// production. Deliberately not prod-locked like the flags above: the
     /// control is reachable everywhere so Listen can be dogfooded in TestFlight.
     ///
-    /// The default follows the build: on for the internal Dev/local builds that
-    /// ship to TestFlight, off for production, so end users still have to opt in
-    /// from the prod debug menu. An explicit toggle in either direction is
-    /// remembered and wins over the default.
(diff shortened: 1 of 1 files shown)


humanagent:
@yewreeka  I'll be OOO next week, feel free tu turn this one if there is a prod merge next week. set it up this pr just in case.

PR #1319, opened 2026-08-13
Add Space share debug action

saulmc:
## Summary

Debug-gated "Share Space" row in the conversation info sheet, behind a new `isSpaceShareEnabled` flag (reachable in every environment; opt-in from the debug menus, including the curated prod menu). Tapping it calls the backend relay `POST /v2/conversations/:conversationId/debug/space-share`, which mints a paste-ready Space import message — one sentence naming the receiving agent's `space-import` skill plus a clone URL carrying a 7-day read-only repository credential — and copies it to the clipboard. Pasting it into another conversation makes that agent rebuild the Space there.

The message embeds a live credential, so it goes straight from the response to the pasteboard: alerts, logs, and error paths never carry it. `ConversationSpaceShareCoordinator` mirrors the existing info-sheet coordinator shape with injectable share + clipboard operations for tests.

One touch on shared plumbing: the strict per-segment URL builder gains a second, non-abilities consumer, so `abilitiesURL` is renamed `endpointURL` and the URL-construction block moves from `+Abilities.swift` to its own `ConvosAPIClient+EndpointURL.swift` — a pure move, no signature changes.

Parallel to convos-client#176 (Android); the Worker endpoint, backend relay, and `space-import` skill land via convos-assistants#3347.

## Why

Internal flow for handing a Space built in one conversation to another. The share is a bounded read grant — the link expires on its own and the receiving agent rebuilds in its own repository, gaining no standing on the source Space.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1319"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789250277&installation_model_id=20076&pr_number=1319&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1319&signature=953b5b53795da3cd3ec39a18bd6f91bd45f966affd4d601831c9c9bd2647b9ce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

> [!NOTE]
> ### Add Space share debug action to conversation info view
> - Adds a `shareSpace(conversationId:variantId:)` API method that POSTs to `/v2/conversations/{id}/debug/space-share`, decodes a `ConvosAPI.SpaceShareLink` on success, and maps backend error codes and timeouts to typed `ConvosAPI.SpaceShareError` cases.
> - Adds `ConversationSpaceShareCoordinator` to manage busy state and alert payloads; on success it copies the returned message to the clipboard and shows a confirmation alert.
> - Adds a "Share Space" button to the debug section of `ConversationInfoView`, visible only when the `isSpaceShareEnabled` feature flag is on; the button shows a spinner and is disabled while the request is in flight.
> - Exposes the feature flag as a toggle in both the internal debug view and the production debug menu.
> - Refactors URL construction into a shared `endpointURL`/`strictPathComponentEncoded` helper (previously duplicated in the abilities endpoint), updating contract tests accordingly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b17682f.</sup>
> 
>

### DIFF
--- a/Convos/Config/FeatureFlags.swift
+++ b/Convos/Config/FeatureFlags.swift
@@ -147,6 +147,22 @@ final class FeatureFlags {
         }
     }
 
+    /// Off by default -- gates the internal action that copies a Space share
+    /// message to the clipboard so another conversation's agent can import
+    /// the Space. Deliberately reachable in every environment; production
+    /// users opt in from the curated prod debug menu.
+    var isSpaceShareEnabled: Bool {
+        get {
+            access(keyPath: \.isSpaceShareEnabled)
+            return UserDefaults.standard.bool(forKey: Constant.spaceShareEnabledKey)
+        }
+        set {
+            withMutation(keyPath: \.isSpaceShareEnabled) {
+                UserDefaults.standard.set(newValue, forKey: Constant.spaceShareEnabledKey)
+            }
+        }
+    }
+
     /// Mock credits/subscription state used by the in-app paywall preview surface
     /// in the Debug menu. Non-production only; defaults to `.plusAmple`.
     var mockCreditsPreset: CreditsStatePreset {
@@ -217,5 +233,6 @@ final class FeatureFlags {
         static let listenParticipationEnabledKey: String = "featureFlags.listenParticipationEnabled"
         static let xmtpBidiStreamsEnabledKey: String = "featureFlags.xmtpBidiStreamsEnabled"
         static let abilitiesV2EnabledKey: String = "featureFlags.abilitiesV2Enabled"
+        static let spaceShareEnabledKey: String = "featureFlags.spaceShareEnabled"
     }
 }
--- a/Convos/Conversation Detail/ConversationInfoView.swift
+++ b/Convos/Conversation Detail/ConversationInfoView.swift
@@ -115,6 +115,7 @@ struct ConversationInfoView: View {
     @State private var showingRestoreInviteTagAlert: Bool = false
     @State private var restoreInviteTagText: String = ""
     @State private var showingLeaveConfirmation: Bool = false
+    @State private var spaceShareCoordinator: ConversationSpaceShareCoordinator = .init()
     @State private var navState: ConversationInfoNavigatorImpl = .init()
     @State private var navigator: ConversationInfoCollector?
 
@@ -658,13 +659,16 @@ struct ConversationInfoView: View {
 // MARK: - Support and debug section
 
 extension ConversationInfoView {
-    // The support rows ship in every environment so production users can
-    // send on-device diagnostics to support; the remaining rows are internal
-    // debugging tools and stay out of production builds.
+    // The support rows ship in every environment so production users can send
+    // on-device diagnostics to support. The Space share row follows its
+    // explicit opt-in flag; the remaining debug rows stay out of production.
     @ViewBuilder
     private var debugInfoSection: some View {
         let isProduction = ConfigManager.shared.currentEnvironment.isProduction
         Section {
+            if FeatureFlags.shared.isSpaceShareEnabled {
+                spaceShareRow
+            }
             if !isProduction {
                 internalDebugRows
             }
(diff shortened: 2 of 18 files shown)


saulmc:
Thanks for the thorough pass — flagging three items that don't hold against the current code, so nobody re-derives them:

**"Must fix: complete `MockAPIClient.shareSpace()`"** — it is complete (`MockAPIClient.swift:135-144`, returns a populated `SpaceShareLink`). The claim is also self-refuting: Unit Tests and Integration Tests both pass on this head, which isn't possible with a compilation failure. The incomplete stubs were in the *test* targets, and they're fixed in this push (see below).

**"Race condition in `isBusy`"** — `ConversationSpaceShareCoordinator` is `@MainActor`, so `share()` is actor-isolated and there is no suspension point between `guard !isBusy` and `isBusy = true`; the first `await` comes after. Two callers can't interleave there. The suggested mitigation (actor-based implementation) is effectively what's already in place.

**"Clipboard failure test"** — `CopyToClipboard` is `(String) -> Void`, non-throwing, so the injected closure can't throw. The test would require changing the signature first.

Taking the accessibility hint suggestion — added.

On the error copy ("check the selected deployment" → "contact support"): keeping the current wording deliberately. This button lives behind an opt-in debug flag, and "the selected deployment" names the environment selector in that same debug menu; routing an internal debug failure to support would send people somewhere that can't help.

Context for reviewers: this PR was rebased from `mpr/agent-desktop` onto `dev`, since the desktop work landed in dev via #1311 (renamed Desktop→Home). That rebase surfaced a real break — dev's ConvosCore unit/integration test split (#1303) means adding a method to `ConvosAPIClientProtocol` requires updating the integration target's `TestStubAPIClientDefaults` and the standalone `ThrowawayPushAPIClient` too. Both are fixed here.

saulmc:
Correction to my previous comment, and a finding worth knowing.

I said the rebase onto `dev` was verified. That was true of everything CI checks — and CI does not check the app target. Building `Convos (Dev)` locally failed:

```
ConversationInfoView.swift:751: error: value of type 'Conversation' has no member 'agentVariant'
```

`Conversation.agentVariant` exists only on `mpr/agent-desktop` (`Storage/Models/Conversation.swift:132`), so this PR did not compile on dev while every check reported green. Fixed by passing `FeatureFlags.shared.effectiveAgentVariantSlug` — the same expression `ConversationView.swift:1463` and `ConversationViewModel.swift:1579` use for the sibling participation calls, and documented as stripped in production by the API client, which matches this PR's existing production-stripping test. `Convos (Dev)` builds clean now.

Two things this surfaced that outlive this PR:

1. **No CI job builds the app target.** The ConvosCore/Keychain/Integration jobs all pass on code the app cannot compile. Any app-target break lands silently.
2. **The app test target does not compile on `dev` today**, independent of this PR: `ConvosTests/ConversationViewModelCapabilityConnectTests.swift:47: error: 'async' call in an autoclosure that does not support concurrency`. That file is byte-identical to dev's copy and untouched here.

Also added the accessibility hint from the review.

saulmc:
Correcting my own comment above: I said no CI job builds the app target. That was wrong, and the real mechanism is worth knowing.

`firebase_pr` (`.github/workflows/firebase-pr.yml` → reusable `xmtplabs/convos-releases/.../ios-firebase-pr.yml`) **does** build the app — it archives and uploads to Firebase App Distribution. It has now run on this PR and passed, which is the external confirmation that the `agentVariant` fix compiles.

Why it never caught the break earlier, two compounding gates:

1. It is filtered `on: pull_request: branches: ["main", "dev"]`. While this PR targeted `mpr/agent-desktop` it was never eligible, so a PR stacked on a feature branch gets **no app-build coverage at all**.
2. Retargeting alone did not trigger it: `gh pr edit --base dev` emits an `edited` event, and the workflow listens for `opened, reopened, synchronize, labeled`. It took the next push. Same reason `SwiftLint` and `check_versions` only appeared on this push.

So the gap is narrower than I described, but the practical consequence for stacked PRs is real: nothing compiles the app until the PR points at `main`/`dev` **and** something pushes afterward.

The other finding stands unchanged — the app test target does not compile on `dev` today (`ConvosTests/ConversationViewModelCapabilityConnectTests.swift:47: 'async' call in an autoclosure that does not support concurrency`), in a file untouched by this PR.

PR #1356, opened 2026-08-19
Show the variant badge in an agent DM

humanagent:
A 1-on-1 with a varianted agent shows no variant marker at all.

The dev variant ribbon rides the agent contact card. A DM deliberately drops that card — `dmItems` nils `group.agentContactCard` and leads with the disclosure header instead — so the ribbon has nothing to attach to. You can pick a variant, open the DM, and have nothing on screen distinguish it from a default agent until its behavior disagrees. That is precisely the failure the variant tooling exists to close, and it's the one surface where it still bites.

### The change

`AgentDmInfoCellView` now takes the agent's `AgentVariantStamp` and renders the shared `AgentVariantRibbon` above the title, so the DM answers "which runtime am I talking to?" the same way the group transcript does, with the same 🧪 label and PR link.

Sized to content and clipped to a capsule rather than the full-bleed bar the contact-card overlay uses — the header is a centered column, where a full-width left-aligned bar would read as a banner across the screen instead of a badge on the agent.

Plumbing: the stamp rides `MessagesListItemType.agentDmInfo` so the cell has it at render time, and the slug is part of the item's `id` so a stamp that lands after first paint (profile sync) re-renders the cell instead of sticking. `AgentDmPageView` reads it off the agent member's profile, where the assistants worker writes it at join.

Dev-only and nil-guarded on both counts: off dev, or for a default agent, the header renders exactly as it does today. The new `init` parameter defaults to nil, so the call site in the share extension is untouched.

### Verification — please read

**I could not build or test this.** I'm running on Linux with no Swift toolchain, so nothing here has been compiled and SwiftLint has not run. That is why it's a draft — CI should be the first thing that compiles it.

Checked by hand instead: every other `agentDmInfo` pattern match is non-binding (`case .agentDmInfo:` in three switches, `if case .agentDmInfo = self`), so adding an associated value leaves them compiling; only the two binding sites are updated. `AgentVariantStamp` is already `Equatable, Hashable, Sendable`, so the enum's synthesized conformances hold. `AgentVariantRibbon` and `ConfigManager` are both already used inside ConvosComposer.

### Related

Backend side landed in convos-assistants#3460 (the join response now reports the applied variant, and the agent row records it); the cache fix that made a fresh variant selectable at all is #1346.

Follow-up worth considering, not in here: the backend now also returns `variantDropped` — the slug you asked for and did not get. Surfacing *that* would catch the case where the pick never bound, which this badge cannot show because there is no stamp to read.

https://claude.ai/code/session_01EQyDQjA1jrkdwPNoG25krY

---
_Generated by [Claude Code](https://claude.ai/code/session_01EQyDQjA1jrkdwPNoG25krY)_

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1356"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789753045&installation_model_id=20076&pr_number=1356&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1356&signature=4651172e6a9fd94eac30200eee45bf456e457f984c0f2f07dcd53c62b77aa8ca"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

> [!NOTE]
> ### Show the agent variant badge in the agent DM info header
> - Adds an optional `variant: AgentVariantStamp?` to the `MessagesListItemType.agentDmInfo` enum case and threads it through to `AgentDmInfoCellView`.
> - In non-production builds, `AgentDmInfoCellView` renders an `AgentVariantRibbon` above the agent name when a non-empty variant stamp is present.
> - The `agentDmInfo` item's computed `id` now includes the variant slug (e.g. `agent-dm-info-none` vs `agent-dm-info-<slug>`), so diffable data sources rebuild the cell when the variant changes.
> - Behavioral Change: all call sites constructing `.agentDmInfo` must now supply the variant argument; the item's identity value has changed from the constant `agent-dm-info`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c0509a4.</sup>
> 
>

### DIFF
--- a/Convos/Conversation Detail/AgentDmPageView.swift
+++ b/Convos/Conversation Detail/AgentDmPageView.swift
@@ -262,7 +262,11 @@ struct AgentDmPageView: View {
                 return item
             }
         }
-        items.insert(.agentDmInfo(agentName: agentName), at: 0)
+        // The stamp lives on the agent member's profile, written by the
+        // assistants worker at join. nil for a default agent, and nil off dev,
+        // where the header renders exactly as before.
+        let variant = dmVm.conversation.members.first(where: { $0.isAgent })?.profile.variant
+        items.insert(.agentDmInfo(agentName: agentName, variant: variant), at: 0)
         return items
     }
 
--- a/ConvosCore/Sources/ConvosComposer/AgentDmInfoCellView.swift
+++ b/ConvosCore/Sources/ConvosComposer/AgentDmInfoCellView.swift
@@ -9,15 +9,33 @@ import SwiftUI
 ///
 /// Figma 7686:39662. The three lines step down the iOS type scale -
 /// .body medium, .subheadline, .footnote - with the disclosure in secondary.
+///
+/// Dev-only: when the agent was built on a variant, the shared variant ribbon
+/// sits above the title as a badge. Everywhere else the ribbon rides the agent
+/// contact card, which a DM deliberately drops - so this header is the only
+/// place a DM can answer "which runtime am I talking to?". Without it a
+/// varianted agent is indistinguishable from a default one until its behavior
+/// disagrees, which is the whole failure mode the variant work exists to close.
 public struct AgentDmInfoCellView: View {
     let agentName: String
+    let variant: AgentVariantStamp?
 
-    public init(agentName: String) {
+    public init(agentName: String, variant: AgentVariantStamp? = nil) {
         self.agentName = agentName
+        self.variant = variant
     }
 
     public var body: some View {
         VStack(spacing: DesignConstants.Spacing.step3x) {
+            if !ConfigManager.shared.currentEnvironment.isProduction, let variant {
+                // Sized to its content and centered, rather than the full-bleed
+                // bar the contact-card overlay uses: this header is a centered
+                // column, and a full-width left-aligned bar would read as a
+                // banner across the screen instead of a badge on the agent.
+                AgentVariantRibbon(variant: variant)
+                    .fixedSize(horizontal: true, vertical: false)
+                    .clipShape(Capsule())
+            }
             Text("1-on-1 with \(agentName)")
                 .font(.body.weight(.medium))
                 .foregroundStyle(.colorTextPrimary)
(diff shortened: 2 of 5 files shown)


humanagent:
Went through the review. Taking one of the four, with reasons for the rest.

**Test coverage — taken.** Added `MessagesListItemTypeIdentityTests.swift` in `c0509a4`. It pins the thing that actually matters: the header's id carries the variant slug, so a stamp landing after first render (profile sync) changes the identity and forces the cell to rebuild. Four cases — no variant, with variant, identity changes when the stamp arrives, and `explainsAnEmptyTranscript` still holds with a variant attached.

**Empty-label guard — already documented.** The comment above it says exactly what the review asks for: the worker forwards `metadata.variant` verbatim for the cosmetic banner (`buildProfileMetadata` in convos-assistants), so a stamp that `parseVariantDescriptor` would reject still decodes here. Empty labels are possible by contract, not a data-quality smell being masked. Skipping the suggested log line — a dev-only cosmetic badge isn't worth an error-level log.

**SwiftLint — not an issue.** `line_length` is 200 in `.swiftlint.yml`. The only line over 150 in the diff is `MessagesListItemType.swift:137` at 180 chars, and it was already 180 at `bb6e33f` — my change added a case name to a list that was long before it. `AgentDmPageView.swift:268` is well under.

**Force-unwrap refactor — declining.** The suggested `shouldShowVariantBadge` computed property needs `AgentVariantRibbon(variant: variant!)`, and `force_unwrapping` is enabled in this repo's SwiftLint config. The `if let` is both shorter and the only form that lints.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1381, opened 2026-08-20
fix(variant): don't let an abandoned pick outlive its flow

humanagent:
Addresses the High finding Macroscope raised on #1380 (merged before it was resolved).

The pick-time binding used `assignIfUnset`, which ignores a nil pick. A prepared conversation outlives a flow that never adopts it — the picker was dismissed, or a scan superseded the create — and goes back in the pool still carrying the slug that flow wrote. The next compose adopts that same conversation, so choosing **"No variant" silently built the agent under the previous variant**, with nothing on screen to explain it.

| scenario | before | after |
|---|---|---|
| pick A → dismiss → compose with no variant | inherits A | cleared |
| pick A → dismiss → compose picking B | inherits A | B |
| pick A → create | A | A |

Every compose now states the pick for the conversation it is about to take, clearing it when there is none. The `.ready` binding keeps `assignIfUnset`, so the pick still wins over it.

On Macroscope's suggested fix — binding only after `consumeUnusedConversationId` returns the claimed conversation — that is the more precise point, but it isn't reachable from the app layer without a shared pending-pick slot: `prepareNewConversation()` claims the conversation and fires `ensureDefaultAgentInConversation` in the same breath (`SessionManager.swift:552`), so the binding would have to be written from inside ConvosCore, which cannot see `FeatureFlags`/`AgentVariantAssignmentStore`. A single pending slot is what `AgentVariantAssignmentStore`'s own doc comment rejects, because two overlapping creation flows race for it. Making the pick-time write authoritative closes the same staleness with no new shared state.

One case this deliberately does not solve: two compose flows overlapping on the same peeked conversation, where the last writer wins. Only one flow adopts it, and that was already true before this change.

Verified `BUILD SUCCEEDED` for `Convos (Dev)`. Two tests added; not executed locally (the test runner clones simulators), so CI is the first run.

https://claude.ai/code/session_01XGnQGVRbGwV8dRe94ZoC6T

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1381"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789844734&installation_model_id=20076&pr_number=1381&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1381&signature=60da9112c41f2fd2d723281916ccb3d24de2ff71323710da81fd18ccbba25e5f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

> [!NOTE]
> ### Bind agent variant at conversation claim time instead of pre-binding to peeked id
> - Moves agent variant binding from a pre-binding step on a peeked prepared conversation id to a callback invoked when the session actually claims a prepared conversation id, so an abandoned pick no longer outlives its flow
> - Adds `SessionManager.claimedConversationVariantBinder` hook, wired in `ConvosApp` to call `AgentVariantAssignmentStore.shared.assign(slug:to:)` on the main actor; passing `nil` clears any prior binding
> - Extends `SessionManager.prepareNewConversation(variantSlug:)` and its protocol/mock implementations to thread the chosen slug through to the binder at claim time
> - `ConversationsViewModel.startNewConversation` no longer pre-binds the variant to a peeked id; it passes the slug directly into `NewConversationViewModel`
> - Behavioral Change: nil `variantSlug` at claim time now clears any existing binding for that conversation id; `AgentVariantAssignmentStore` is expected to handle replacement and clearing semantics
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e05087e.</sup>
> 
>

### DIFF
--- a/Convos/Conversation Creation/NewConversationViewModel.swift
+++ b/Convos/Conversation Creation/NewConversationViewModel.swift
@@ -524,7 +524,7 @@ class NewConversationViewModel: Identifiable, Hashable {
 
             switch mode {
             case .newConversation, .newAgent, .newConversationWithTemplate:
-                let (messagingService, existingConversationId) = await session.prepareNewConversation()
+                let (messagingService, existingConversationId) = await session.prepareNewConversation(variantSlug: agentVariantSlug)
                 guard !Task.isCancelled else { return }
                 let inboxElapsed = (CFAbsoluteTimeGetCurrent() - perfStartTime) * 1000
                 Log.info("[PERF] NewConversation.inboxAcquired: \(String(format: "%.0f", inboxElapsed))ms")
@@ -554,7 +554,7 @@ class NewConversationViewModel: Identifiable, Hashable {
                 )
 
             case .newConversationWithMembers(let initialMemberInboxIds, _):
-                let (messagingService, existingConversationId) = await session.prepareNewConversation()
+                let (messagingService, existingConversationId) = await session.prepareNewConversation(variantSlug: agentVariantSlug)
                 guard !Task.isCancelled else { return }
                 let inboxElapsed = (CFAbsoluteTimeGetCurrent() - perfStartTime) * 1000
                 Log.info("[PERF] NewConversation.inboxAcquired: \(String(format: "%.0f", inboxElapsed))ms")
--- a/Convos/Conversations List/ConversationsViewModel.swift
+++ b/Convos/Conversations List/ConversationsViewModel.swift
@@ -523,29 +523,6 @@ final class ConversationsViewModel {
     /// `onStartConvo` so the variant picker can run ahead of it and then
     /// continue into the same flow.
     func startNewConversation(agentVariantSlug: String? = nil) {
-        // Bind the pick to the conversation this flow is about to adopt,
-        // before the flow starts. The warm cache hands out a conversation it
-        // prepared earlier, and that conversation's default agent is
-        // provisioned the moment it is adopted — concurrently with the
-        // `.ready` handler that would otherwise write this binding. Writing it
-        // here means the assignment is already on disk whenever the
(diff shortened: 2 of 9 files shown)


humanagent:
Macroscope was right, and I verified the divergence rather than arguing it: `peek` returns `preparedIdBox.value` (the id prepared in this session) while `consumeUnusedConversationId` runs its own query over every eligible row with no `ORDER BY`. Two real ways to diverge — more than one unused row present, and app restart, where the box is nil but the claim still succeeds and the pick goes unwritten, leaving the original race open.

Rewritten in `e05087e5`: the pick travels into `prepareNewConversation(variantSlug:)` and is recorded against the conversation the claim actually returned, before the claim-time provision can read it. The `peek` guess is gone.

On why the binding is written inside ConvosCore rather than after the call returns: `prepareNewConversation` fires `ensureDefaultAgentInConversation` in the same breath as the claim, so binding on return would still race it. The slug travels per call through a hook the app installs, so no shared pending-pick slot is introduced.

PR #1404, opened 2026-08-22
DNM - Paused Composer

onefovrth:
## What

When an agent's participation is set to **Pause**, the agent DM composer now blocks sending instead of silently dropping messages the agent will never see.

In the **agent DM** (group composer and Share extension are untouched):
- Send button becomes a `pause.fill` glyph in the disabled/minimal visual, but stays tappable.
- Composer hint text reads **"{agent} is paused"** (per design) instead of "Chat with {agent}".
- Every send path is blocked: send button, return key (`.onSubmit`), and voice-memo entry points (empty-composer swap + the `+` menu's voice-note row, now disabled).
- A blocked attempt raises an alert — **"Agent is paused" / "{agent} is paused and won't see new messages right now."** — with **Unpause** (resumes to speak-freely) and **Not now**.

## How

- `sendButtonPaused` + `onPausedSendTap` params thread through `MessagesView` → `MessagesBottomBar` → `MessagesInputView`, all with safe defaults (`false` / no-op), so the group and Share paths are unchanged.
- `AgentDmPageView` receives `sendButtonPaused` + `onUnpauseAgent`, owns the alert, and derives the paused placeholder.
- `ConversationView.agentPage` sources both from its optimistic `AgentParticipationStore` (`participation?.level == .paused`, `set(.speakFreely)` for unpause) — instant on local taps, synced for remote changes, and naturally gated on the participation feature flag.

## Notes for review

- Unpause always resumes to **Chat (speak freely)**; the prior non-paused mode (e.g. Listen) isn't restored, since the store doesn't track it.
- Text field stays editable while paused — only sending is blocked.
- Edge case left as-is: if pause flips while a voice memo is already recorded, the review sheet's send is still live.

## Testing

- `xcodebuild build` (Convos (Dev)) succeeds; SwiftLint `--strict` clean.
- Manual (feature flag on, conversation with an agent): pause on the group tab → Agent tab shows the paused glyph + hint; tapping raises the alert; Unpause restores sending; `+` menu voice-note disabled while paused.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1404"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789956538&installation_model_id=20076&pr_number=1404&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1404&signature=9caf6172a4de9449237954a5e69e568a4391bb2503c0af6e93f7d47b10715deb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

> [!NOTE]
> ### Add paused-agent composer state and UIKit unpause alert to `AgentDmPageView`
> - When an agent's participation level is `.paused`, the composer shows a pause-glyph send button with tertiary styling and a placeholder reading "<agent> is paused"; voice memo entry is suppressed.
> - Send attempts (tap or Return key) trigger a new `PausedAgentAlert` — a `UIViewControllerRepresentable` wrapping a `UIAlertController` pinned to light appearance — offering OK and Unpause actions instead of sending.
> - Tapping Unpause in [ConversationView.swift](https://github.com/xmtplabs/convos-ios/pull/1404/files#diff-315e38ca52e75c79370579c65f3a218ca32afe08c50f49e8165bc2d562a1ce58) asynchronously sets participation from `.paused` to `.mentionsOnly` when the device is not read-only.
> - The paused state and tap handler are threaded from `AgentDmPageView` through `MessagesView`, `MessagesBottomBar`, and `MessagesInputView` via new `sendButtonPaused` and `onPausedSendTap` parameters.
> - Risk: the previous SwiftUI `.alert` modifier in `AgentDmPageView` is removed in favor of the UIKit-based `PausedAgentAlert`; any logic that relied on the SwiftUI alert presentation path will no longer fire.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 145c6c6.</sup>
> 
>

### DIFF
--- a/Convos/Conversation Detail/AgentDmPageView.swift
+++ b/Convos/Conversation Detail/AgentDmPageView.swift
@@ -28,6 +28,14 @@ struct AgentDmPageView: View {
     /// Mirrors ConversationView's effectiveReadOnly: a removed or stale
     /// device must not be able to send into agent DMs.
     let isReadOnly: Bool
+    /// True when the conversation's agent participation is paused. The DM's send
+    /// button then shows the paused visual, sends are blocked, and the composer
+    /// hint changes; a blocked attempt raises the paused alert.
+    var sendButtonPaused: Bool = false
+    /// Resumes the agent (participation -> speak freely) from the paused alert's
+    /// Unpause action. Owned by ConversationView, which holds the participation
+    /// store the DM page can't reach.
+    var onUnpauseAgent: () -> Void = {}
     /// True while the Agent tab is the selected tab. The backing views stay
     /// mounted across tab switches, so activation drives the read state and
     /// the active-DM push lane. (Composer focus transfers are owned by
@@ -61,6 +69,7 @@ struct AgentDmPageView: View {
     private var agentName: String { session.agentName }
 
     @State private var emptyStateSettled: Bool = false
+    @State private var showingPausedAgentAlert: Bool = false
 
     /// Centred over the page and faded rather than inserted into the list, so
     /// it never holds a scroll position. Never takes hits - it is only text.
@@ -300,6 +309,9 @@ struct AgentDmPageView: View {
 
     private func dmMessagesView(_ dmVm: ConversationViewModel) -> some View {
         @Bindable var dmVm = dmVm
+        let messagePlaceholder: String = sendButtonPaused
+            ? "\(agentName) is paused"
+            : "Chat with \(agentName)"
         return MessagesView(
             contextMenuState: contextMenuState,
             conversation: dmVm.conversation,
@@ -313,7 +325,7 @@ struct AgentDmPageView: View {
             conversationImage: $dmVm.conversationImage,
             displayName: $dmVm.myProfileViewModel.editingDisplayName,
             messageText: $dmVm.messageText,
-            messagePlaceholder: "Chat with \(agentName)",
+            messagePlaceholder: messagePlaceholder,
             pendingMediaAttachments: dmVm.pendingMediaAttachments,
             composerLinkPreview: dmVm.pastedLinkPreview,
             pendingInviteURL: dmVm.pendingInvite?.fullURL,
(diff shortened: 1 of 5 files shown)


onefovrth:
Addressed the critical finding in 6e80e4e7.

**Read-only unpause (required fix):** `onUnpauseAgent` is now gated on `!effectiveReadOnly` at its source in `ConversationView.agentPage`, the same gate the composer's sends use, so a removed/stale device can't flip participation back to speak-freely via the alert's Unpause button.

```swift
onUnpauseAgent: {
    guard !effectiveReadOnly, let participation else { return }
    Task { await participation.set(.speakFreely) }
},
```
(The closure lives in ConversationView rather than AgentDmPageView, since `effectiveReadOnly` + the participation store are owned there; the DM page already receives `isReadOnly: effectiveReadOnly`.)

**Copy:** the longer wording ("Paused agents can't receive 1:1 messages nor see anything new in the groupchat") is intentional and design-approved — it supersedes the shorter string in the original PR description.

**Recorded-memo-then-pause race:** confirmed acceptable as-is (rare; doesn't justify the added complexity).

**Tests:** these are UI-only SwiftUI/UIKit surfaces; unit coverage is a noted follow-up gap rather than a blocker.

yewreeka (approved):

PR #1410, opened 2026-08-23
Agent relay: ConvosCore client and store

lourou:
## Summary

Adds the ConvosCore side of external agent relay: chatting from Convos with
an AI agent the user already runs on another platform (Town or Tasklet),
and copying its answers into a conversation as an editable draft.

This PR is the client and storage layer only, with no UI:

- `AgentRelayClient` and `AgentRelayBackendAPI`: create a relay request,
  trigger the agent's webhook directly from the device, long-poll for the
  result, and acknowledge it.
- `AgentWebhookTransport` and `WebhookURLValidator`: send the webhook call
  and validate the connection URL the user configures.
- `AgentConnectionStore`, `AgentChatDatabase`/`AgentChatMigrations`,
  `AgentChatRepository`, `AgentChatWriter`, `AgentTurn`: a separate GRDB
  database for the agent chat transcript and stored connection, kept apart
  from the main conversations database since this data is device-local and
  does not roam.
- `AgentRelayRecoveryCoordinator`: resumes pending turns and reconciles
  completed-but-unacked ones on launch/foreground.
- `AgentHistoryBuilder`, `AgentRelayPushPayload`, `AgentRelayReset`,
  `AgentRelayTypes`, `ExternalAgentProvider`: history windowing for
  stateless agent runs, the completion-push payload shape, and the
  provider abstraction (Town and Tasklet today; open to others later).

## Why

Prompts, conversation history, and the user's webhook secret should never
transit or rest on Convos servers — the phone talks to the agent platform
directly, and the backend only runs the return path. Keeping that logic in
ConvosCore, tested against protocol seams and a stubbed `URLSession` layer,
lets it be verified without a live backend or a real agent account.

This is the base of the stack in this repo. The app UI on top of it
(agent chat screens, setup flows, and notification handling) is the next
PR in the stack — merge this one first. This PR is client-only and has no
user-visible effect on its own; end-to-end, none of it works until the
backend module (a separate repository) has been merged and deployed, since
until then the app would be calling endpoints that do not exist.

## Open questions

- Not yet verified against real accounts: whether Town and Tasklet accept
  a custom remote MCP server with no auth as the return path, and whether
  a routine/agent run remembers previous runs (whether the `history` this
  client sends is load-bearing or just a convenience).
- `PRODUCT.md` is currently untracked in this repo, so the copy rules it
  is meant to hold are not readable from a clean checkout. Worth fixing
  before agent chat copy is finalized.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1410"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790080497&installation_model_id=20076&pr_number=1410&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1410&signature=212274497b4be5ae8b44aee21c0c8c585028aa74d04ffc4490b4e415d51ec65a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

> [!NOTE]
> ### Add Agent Relay client, store, and backend API to ConvosCore
> - Introduces the Agent Relay feature: minting relay requests via backend API, triggering external provider webhooks, long-polling for completion, and persisting turns in a new GRDB-backed `agent-chat.sqlite` database.
> - Adds `AgentRelayClient` as the high-level orchestrator for a relay turn, plus `AgentRelayRecoveryCoordinator` to reconcile local state with backend on app launch/foreground.
> - Adds `AgentConnectionStore` to persist and validate per-provider connections (bearer secret in Keychain for `.town`, capability URL for `.tasklet`) and `WebhookURLValidator` to block private/local/obfuscated URLs including DNS resolution checks.
> - Adds `PendingComposerDraftStore` for staging composer drafts with a 10-minute TTL, `AgentRelayReset.wipeAll` for account-level cleanup, and `AgentRelayPushPayload` for parsing APNs notifications.
> - Extends `ConvosAPIClientProtocol` with `authorizedRequest(for:method:queryParameters:)` to build authenticated `URLRequest`s outside the client's session; `SessionManager` account reset now calls `AgentRelayReset.wipeAll` and throws on failure.
> - Risk: `ConvosAPIClientProtocol` gains a new required method (`authorizedRequest`); out-of-tree conformers must implement it. `SessionManager` reset path now throws if agent relay wipe fails, which may break callers not expecting new error cases.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized df10d11.</sup>
> 
>

### DIFF
--- a/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
+++ b/ConvosCore/Sources/ConvosCore/API/ConvosAPIClient.swift
@@ -23,6 +23,14 @@ public protocol ConvosAPIClientProtocol: AnyObject, Sendable {
                  method: String,
                  queryParameters: [String: String]?) throws -> URLRequest
 
+    /// Builds a request with the JWT attached (override token first, then
+    /// the keychain slots) and returns it for the caller to execute on its
+    /// own session. The agent relay uses this so long-poll responses never
+    /// pass through the client's body-logging path.
+    func authorizedRequest(for endpoint: String,
+                           method: String,
+                           queryParameters: [String: String]?) async throws -> URLRequest
+
     /// Register device with AppCheck authentication (no JWT required - device-level operation)
     func registerDevice(deviceId: String, pushToken: String?) async throws
 
@@ -677,6 +685,12 @@ final class ConvosAPIClient: ConvosAPIClientProtocol, Sendable {
     ) throws -> URLRequest {
         let request = try request(for: path, method: method, queryParameters: queryParameters)
         return attachingAuthHeader(to: request)
+    }
+
+    func authorizedRequest(for endpoint: String,
+                           method: String,
+                           queryParameters: [String: String]?) async throws -> URLRequest {
+        try authenticatedRequest(for: endpoint, method: method, queryParameters: queryParameters)
     }
 
     /// Attaches the auth header to a prebuilt request. Split from
--- a/ConvosCore/Sources/ConvosCore/API/MockAPIClient.swift
+++ b/ConvosCore/Sources/ConvosCore/API/MockAPIClient.swift
@@ -24,6 +24,13 @@ final class MockAPIClient: ConvosAPIClientProtocol, Sendable {
         return URLRequest(url: url)
     }
 
+    func authorizedRequest(for endpoint: String, method: String, queryParameters: [String: String]?) async throws -> URLRequest {
+        var request = try request(for: endpoint, method: method, queryParameters: queryParameters)
+        request.httpMethod = method
+        request.setValue(overrideJWTToken ?? "mock-jwt-token", forHTTPHeaderField: "X-Convos-AuthToken")
+        return request
+    }
+
     func registerDevice(deviceId: String, pushToken: String?) async throws {
         // Mock implementation - no-op
     }
--- a/ConvosCore/Sources/ConvosCore/AgentRelay/AgentChatDatabase.swift
+++ b/ConvosCore/Sources/ConvosCore/AgentRelay/AgentChatDatabase.swift
@@ -1,0 +1,60 @@
+import Foundation
+import GRDB
+import SQLite3
+
+/// Owns the `agent-chat.sqlite` pool in the app group container. Opened by
+/// the main app and by the NotificationService extension, so the pool uses
+/// WAL with the persist-WAL file control and a busy timeout, mirroring the
+/// conversation database without sharing its migrator.
+public final class AgentChatDatabase: Sendable {
+    public let pool: DatabasePool
+
+    public init(environment: AppEnvironment) throws {
+        let directory = environment.defaultDatabasesDirectoryURL
+        try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true)
+        let url = directory.appendingPathComponent(Constant.fileName)
+        pool = try Self.openPool(at: url.path)
+    }
+
+    /// A throwaway database for tests: a unique temporary file, because a
+    /// `DatabasePool` cannot be opened on an in-memory database.
+    public static func inMemoryForTests() throws -> AgentChatDatabase {
+        let url = FileManager.default.temporaryDirectory
+            .appendingPathComponent("agent-chat-\(UUID().uuidString).sqlite")
+        return try AgentChatDatabase(pool: openPool(at: url.path))
+    }
+
+    private init(pool: DatabasePool) {
+        self.pool = pool
+    }
+
+    private static func openPool(at path: String) throws -> DatabasePool {
+        var config = Configuration()
+        let isNSE = Bundle.main.bundleIdentifier?.contains("NotificationService") ?? false
+        config.label = isNSE ? "AgentChatDB-NSE" : "AgentChatDB-MainApp"
+        config.foreignKeysEnabled = true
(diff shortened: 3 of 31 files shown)


lourou:
Pushed a fix for all eight findings in `682ca515`:

1. `AgentRelayReset.wipeAll` (High) - each cleanup step (chat DB wipe, town credential, tasklet credential) now runs in its own do/catch; the first error is captured and rethrown only after every step has been attempted, so a failure on one no longer skips the rest.
2. `WebhookURLValidator` trailing-dot bypass (High) - strips one terminal DNS root dot before the `localhost`/`.local` checks, so `https://localhost./webhook` and `https://printer.local./webhook` are rejected.
3. `AgentRelayRecoveryCoordinator` silent data loss (High) - a listed completion with no local row now goes through `client.collect(...)` (fetch, persist, ack) instead of a bare ack, so the result is saved before the mailbox is deleted.
4. `WebhookURLValidator` DNS rebinding (High) - added an injectable resolver (default: `getaddrinfo`) that resolves the hostname at validation time and rejects it if any resolved address is loopback, private, link-local, multicast, or unspecified. This closes the validation-time hole; it does not pin the resolved address into the later request, so the address could still change between validation and send - that's a follow-up, noted in the code.
5. `SessionManager.tearDownInbox` (High) - now captures the relay wipe error, still clears the cached service reference, then rethrows, so `deleteAllInboxesWithProgress()` correctly reports failure instead of swallowing it. Safe to rethrow now that #1 guarantees every cleanup step was attempted first.
6. `AgentHistoryBuilder` budget trim (Medium) - trims complete user/agent pairs instead of a single oldest entry, bounds-checked so it can't crash on an odd count or a single oversized pair.
7. `AgentRelayRecoveryCoordinator` listing failure (High) - a failed `completedEntries()` call now returns immediately instead of falling through with an empty list, so a transient failure can no longer cause a locally-pending turn to be marked expired.
8. `AgentRelayRecoveryCoordinator` terminal-state reconciliation (High) - local rows already marked `.failed`/`.expired`/`.collectedElsewhere` are now reconciled against a listed completion via `collect`, same as an untracked row. Required loosening `AgentChatWriter.markCompleted`'s update guard from `status == .pending` to `status != .completed`, since the old guard silently no-op'd on any non-pending existing row.

64 AgentRelay tests pass (10 new), `swift build --build-tests`, `swiftlint --strict` on all touched files, and `git diff --check` are all clean.

`feature/agent-relay-ios` will need a rebase onto this commit.

lourou:
Addressed the follow-up finding on webhook host validation: added RFC 6598 shared address space (`100.64.0.0/10`, the Tailscale/CGNAT range) to the blocked IPv4 ranges, plus the adjacent `240.0.0.0/4` reserved/broadcast tail, applied to both literal and DNS-resolved addresses since they share the same predicate. Fixed in 7efe246a.

lourou:
Fixed the three new findings in f6c8363:

- `watch` deadline overshoot: `waitMs` is now capped to the remaining time each loop iteration, with a final zero-wait fetch right at the deadline instead of skipping straight to `stillWorking` unchecked.
- `AgentConnectionStore.delete`: provider defaults and `activeProvider` are now cleared unconditionally even when the Keychain delete throws, then the original error is rethrown - same fix shape as the earlier `wipeAll` hardening.
- SSRF range table: did the full pass instead of another single-range patch. `isBlockedIPv4`/`isBlockedIPv6` are now table-driven (network + prefix length, one prefix-match helper) covering the full RFC list including `198.18.0.0/15`, `fec0::/10`, and IPv4-mapped addresses delegating to the IPv4 table.

On the still-open DNS rebinding finding (`WebhookURLValidator.swift:135`): validation-time resolution from 682ca51 is intact and closes the hole at connect time. Pinning the resolved address at request time needs a custom `URLProtocol` and is a follow-up, not part of this pass.

71 tests passing (was 68), swiftlint clean, no merge conflicts with dev.

lourou:
Fixed both in 9d0d5c3e:

- `send` now derives the provider from the connection instead of taking it as a separate parameter, so mint/journal/trigger can no longer disagree on which endpoint gets the prompt, history, and return token.
- `collect` no longer falls back to `.town` when the provider is unknown and there's no local row; it leaves the completion in the mailbox and logs, so it can be picked up correctly once attributable.

lourou:
Fixed the remaining town guess in df10d114:

- `watch`'s completion branch now resolves the provider from the journaled row (same lookup `collect` uses) instead of hardcoding `.town`. If it can't be attributed, it logs a warning and leaves the turn unacked rather than persisting under a guessed provider.

PR #1411, opened 2026-08-23
Agent relay: app surfaces

lourou:
## Summary

Adds the app surfaces for external agent relay on top of the ConvosCore
client: the agent chat screen, the Town and Tasklet connection setup
flows, copying a conversation message into the agent composer and copying
an agent reply into a conversation as an editable draft, App Settings entry
points, a feature flag, and the notification-service-extension handling
for the completion push.

Notable pieces:
- `AgentsHomeView`/`AgentsHomeMode`/`AgentsUnavailableView`: the
  account-level agent chat, reachable from App Settings.
- `TownSetupView`/`TaskletSetupView`: connection setup, generating the
  webhook instructions for the user to paste into their own routine/agent.
- `ConversationPickerView`/`ConversationPickerMode`: the "copy to convo"
  destination picker.
- `ConversationViewModel+PendingDraft`, changes to `MessagesInputView` and
  the message context menu: pre-filling an editable draft in either
  direction, nothing auto-sends.
- `AgentRelayNotificationHandler` and `NotificationService.swift`: the NSE
  path that fetches the result with `wait_ms=0`, writes it to the local
  store, and acks on a completion push.
- `FeatureFlags.swift`: gates the feature.

## Why

This is the user-facing half of the feature described in the design doc:
a private, account-level chat with an agent the user already runs
elsewhere, with control staying entirely on the user's side — nothing is
posted anywhere automatically.

Stacked on #1410 (the ConvosCore client and store) — merge that first.
End-to-end,
this also needs the backend module (a separate repository) merged and
deployed first; until then the relay endpoints the app calls do not exist.

## Open questions

- Not yet verified against real accounts: whether Town and Tasklet accept
  a custom remote MCP server with no auth as the return path, and whether
  a routine/agent run remembers previous runs.
- The completion-push path (NSE fetch, local write, ack) has not been
  proven end-to-end on a simulator, which has no notification service
  extension; it needs a physical device to validate.
- `PRODUCT.md` is currently untracked in this repo, so the copy rules it
  is meant to hold are not readable from a clean checkout. The agent chat
  and setup copy in this PR was written without that reference.

---
<a href="https://app.blacksmith.sh/xmtplabs/codesmith/convos-ios/pr/1411"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790080503&installation_model_id=20076&pr_number=1411&repository=xmtplabs%2Fconvos-ios&return_to=https%3A%2F%2Fgithub.com%2Fxmtplabs%2Fconvos-ios%2Fpull%2F1411&signature=f0198023dfadd360832dd45df3e0516f3a38bb49c6fef2265a17bb806e061758"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

> [!NOTE]
> ### Add agent relay app surfaces: Agents tab, chat UI, setup, and notification handling
> - Introduces a new `agents`/Relay tab behind `FeatureFlags.shared.agentRelayEnabled`, routing to [AgentsHomeView](https://github.com/xmtplabs/convos-ios/pull/1411/files#diff-a4de37d5303e2c173eb0c5511cd3b2bd23474508e0b4ba495c31c685e3ab72f8) or provider setup screens for Tasklet and Town providers
> - Adds the full agent chat experience: [AgentChatView](https://github.com/xmtplabs/convos-ios/pull/1411/files#diff-df4e2dba5258ffd729843eadfd3ed0782c83adf42025513cf34b4f6b0ef63da8), [AgentChatViewModel](https://github.com/xmtplabs/convos-ios/pull/1411/files#diff-5205680ee81ab68d576ef8ed79afcfa83b4d60338a5f05cf878534a8b489ebe8), chat bubbles, link confirmation, and a conversation picker for copying replies into convos
> - Wires `AgentRelayDependencies` for end-to-end relay lifecycle: foreground push collection, launch/foreground recovery, notification cleanup, and SwiftUI environment injection in [ConvosApp](https://github.com/xmtplabs/convos-ios/pull/1411/files#diff-de23c5a6e10cb84b89d91ed62b707b654c28566f38cc339af8b3cbbf6e003014)
> - Extends `AgentRelayClient` with bounded exponential backoff on fetch failures, a `superseded` turn state, `waitUntilReady` gating, and `notFound`→`collectedElsewhere` handling; notification service now collects relay results in-extension with a 15s timeout and falls back to original content on timeout
> - Risk: `ConversationsViewModel.selectConversation(id:)` now clears `selectedInitialAgentDmInboxId` instead of preselecting the most recent unread agent DM; verify callers in [ConversationsViewModel](https://github.com/xmtplabs/convos-ios/pull/1411/files#diff-e9ac0abeb6f4ab9d1335cf93b09f18a543936af17c742b0f60b5f57e26c07fce) still route correctly
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0607c12.</sup>
> 
>


lourou:
Rebased onto `feature/agent-relay-core`'s latest tip (`7efe246a`, which now includes both the eight Macroscope fixes in `682ca515` and the RFC 6598 webhook fix that landed just after). Pushed the app-side Macroscope fixes in `1218e1e9`:

1. `AgentChatViewModel.swift:254` (Medium) - added provider-scoped repository queries (`turns(provider:limit:)` / `turnsStream(provider:limit:)`, additive on `AgentChatRepositoryProtocol`) so the turn limit is applied after filtering by provider instead of before. A provider with older turns no longer shows an empty/truncated transcript because another provider has 200 newer rows.
2. `MainTabView.swift:211` (Medium) - clear `conversationsViewModel.selectedConversationId` before switching to the Agents tab, so a Chats selection can no longer suppress the Agents nav/tab bars or leave a stale conversation indicator rendering.
3. `AgentRelayRecoveryCoordinator.swift:34` (Medium) - not applied: conflicts with #1410's fix, which takes precedence because expiring is irreversible. `recover()` still returns early when the completed-entries listing fails rather than falling through with an empty list.
4. `AgentChatViewModel.swift:118` / `AgentChatWriter.swift:108` (Medium) - the `checkAgain` follow-up guard now permits `.superseded` in addition to `.pending`, so "Check again" on a superseded turn resumes waiting instead of dead-ending. No writer-side change was needed; `markCompleted`'s guard already accepts superseded rows from #1410's fix.

77 AgentRelay tests and the 7 `AgentChatViewModelTests` pass, including new regression coverage for both applied fixes.

lourou:
Rebased onto the latest `feature/agent-relay-core` (watch deadline cap, connection-store cleanup, table-driven SSRF list) and pushed a follow-up fix commit (`f3fd85d7`) addressing this round's findings:

- Copy-to-conversation dead-end: the picker's selection now resolves the conversation by id (same path deep links use) instead of relying on the main list's currently loaded page, so it no longer stalls when the target isn't loaded yet.
- Clear history orphaning mailboxes: `clearHistory` now walks every settled turn for the provider (not just the capped 200-row UI list), acknowledges each live mailbox before deleting its row, and keeps any row whose ack fails so it can retry later instead of silently discarding the error.
- Superseded turns never settling: `markExpired` and `markCollectedElsewhere` now also transition `.superseded` rows, so a "check again" that comes back expired/not-found actually settles instead of looping forever.
- Foreground push suppression too coarse: visibility is now tracked per provider instead of one global flag, so a Tasklet completion push is no longer swallowed just because Town's chat happens to be open.
- Whitespace-only composer text: Stop now uses the same whitespace-aware emptiness check as the Send button, so whitespace no longer shows a disabled Send instead of Stop.
- Copy to agent ignoring the feature flag: the context-menu guard now checks `agentRelayEnabled` too, so it's hidden consistently with the tab and settings entry points when the flag is off.
- Mock API hitting the network: `MockAPIClient` now takes an injectable request executor with deterministic canned relay responses (mint/poll/ack) instead of making a real request to example.com.

On the recovery-coordinator thread (`entries = []` on a listing failure): leaving that as-is on purpose. It conflicts with the earlier fix in #1410 that deliberately returns early on a listing failure - expiring local turns off a transient listing failure is irreversible, while skipping one recovery pass is free since it just retries on the next launch or foreground. Not applying that suggestion.

Verified post-rebase: ConvosCore package build + full AgentRelay test suite (82/82) green, SwiftLint clean on touched files, Dev build + targeted app test suites green, and a clean merge-tree against `dev`.

lourou:
Fixed in 42c59987, rebased on the core provider fix:

- Clear history now keeps superseded turns instead of deleting them - superseded turns are kept because they are still running on the provider, and acking them would close a live mailbox and lose the eventual answer. Settled now means completed/failed/expired/collected-elsewhere only.
- Conversation picker: no code change needed. Selection already resolves out-of-window ids end to end (picker -> notification -> `selectConversation(id:)` -> `resolveConversation(id:)` -> pager fetch), so this thread can close.
- Group notification tap no longer reopens the agent-DM page: `selectConversation(id:)` clears `selectedInitialAgentDmInboxId` for ordinary selections; the explicit agent-DM notification path sets it independently and is untouched.
- A transient Keychain read failure during Discover reload no longer wipes connected-provider state; it now keeps the previous state and logs the error.
- Applying a staged composer draft no longer gets swallowed by paste-link detection when the draft contains a URL.

One more town guess was also cleared out of the core watch path (df10d114 on #1410); `grep -n "\.town"` under AgentRelay now only matches legitimate enum cases.

Not in this pass: NotificationService/AgentRelayNotificationHandler.swift:16 (collect returning nil on an unattributable result blanks the notification instead of showing it) is a separate, real finding - left open.

lourou:
Fixed the empty-notification regression: when the extension can't attribute a collected result it now falls back to the original push instead of showing nothing (64026a6).
Added a pure content-selector function in ConvosCore with a unit test covering the nil case.

lourou:
Fixed all four in 8767eb4:

- Cold-launch agent notification tap: deferred the `NotificationCenter` post with `DispatchQueue.main.async`, matching the existing conversation and explosion tap paths.
- Failed reconnect clobbering credentials: `AgentSetupViewModel.connect` now captures the previous connection and active provider before saving, and restores them (or deletes the newly-saved bad credentials if there was no previous connection) when the round trip throws or returns `.failed`/`.expired`.
- Staged draft losing a URL: `applyPendingComposerDraft` now syncs the paste-detection baseline right after assigning the draft text, so the next keystroke isn't misread as pasting the whole draft. Added the regression test for this one.
- `AgentChatViewModel.submit()` no longer clears the composer or supersedes the pending turn until the connection and session-readiness checks pass, so a failed submit leaves the draft and the pending turn intact.

Added 4 tests covering the above, all green locally along with lint, merge-tree against dev, and a targeted build/test pass on device sim.

lourou:
Fixed in adf05f22: the composer clear after `waitUntilReady()` now only fires if the text still matches the submitted prompt, so typing a new message while the session becomes ready no longer wipes it out.
Superseding the prior in-flight turn still happens unconditionally; added a regression test that suspends readiness, edits the composer mid-flight, and asserts both the text survives and the supersede/mint proceed.

lourou:
Renamed the Agents tab (and its matching Settings row / header) to Connect - enum case and types stay as-is, just the visible label.

Relabeled the "Agent relay" debug toggle to "Enable Connect (BYOA)" and moved it to sit right under the Connections v2 controls in both debug menus.

Fixed the preview-backend note so it's keyed on the actual backend host (per-PR preview deploys only) instead of the app's own bundle id, which was flagging PR builds pointed at the shared dev backend.

5331545f

lourou:
Renamed the surface to **Relay** and tightened the copy (0607c12c).

- Tab label: "Connect" -> "Relay" (tab icon also swapped from the node-network glyph to `antenna.radiowaves.left.and.right`, a clearer relay/signal metaphor and visually distinct from the other two tabs)
- Debug flag label: "Enable Connect (BYOA)" -> "Enable Relay (BYOA)" (flag key unchanged)
- App Settings row: "Connect" -> "Relay" (and its unavailable-state fallback: "Agents unavailable" -> "Relay unavailable")
- Surface header and unavailable-state title: "Connect" -> "Relay", "Agents are unavailable" -> "Relay is unavailable"
- Empty-state intro: "Connect an agent you already run somewhere else." -> "Relay work to an agent you already run somewhere else." (kept the second sentence as-is)
- Kept "Connect" wherever it's the actual action of connecting a provider (Connect Town, Connect Tasklet, the Connect buttons, "Connect this agent first.")
- Notification copy no longer promises delivery: "You will get a notification when it replies" -> "We'll notify you when it replies - or check back here"; same softening on the intro notification line
- "conversation" -> "convo" in the credential note
- Standardized "transcript"/"messages" -> "chats" in the on-device-storage and clear-history copy

SwiftLint clean, build succeeds, no conflicts with dev.

